PR #8, opened 2024-07-09
fix: pre-commit syntax in Makefile

tskir:
Since we are using Poetry for dependency and virtual environment management, trying to source .venv/bin/activate results in an error. I have replaced those three occurrences with `poetry run`.


DSuveges (approved):
All makes sense.

PR #10, opened 2024-07-09
feat: finemapping template and DAG for UKB PPP

tskir:
The idea is to have a common finemapping _template,_ which specific DAGs can reuse and modify according to their needs.

### DIFF
--- a/src/ot_orchestration/dags/ukb_ppp_finemapping.py
+++ b/src/ot_orchestration/dags/ukb_ppp_finemapping.py
@@ -5,13 +5,14 @@ from __future__ import annotations
 import time
 from pathlib import Path
 
-import common_airflow as common
 from airflow.decorators import task
 from airflow.models.dag import DAG
 from airflow.providers.google.cloud.operators.cloud_batch import (
     CloudBatchSubmitJobOperator,
 )
-from templates.finemapping import finemapping_batch_job
+
+from ot_orchestration.templates.finemapping import finemapping_batch_job
+from ot_orchestration.utils import common
 
 STUDY_LOCUS_BASE_PATH = (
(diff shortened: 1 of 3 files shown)


project-defiant (approved):

tskir:
@project-defiant This has been quite substantially rewritten compared to the first draft. Could you do another round of reviews, please?

src/ot_orchestration/dags/ukb_ppp_finemapping.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+with DAG(
+    dag_id=Path(__file__).stem,
+    description="Open Targets Genetics — finemap study loci with SuSie",
+    default_args=common.shared_dag_args,
+    **common.shared_dag_kwargs,
+) as dag:
+    (
+        FinemappingBatchOperator.partial(
```
project-defiant, line 41: The partial will not work beyond the threshold, and I have tested it on local airflow DAG, this breaks on around ~5k partial tasks even with the threshold increase.

https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dynamic-task-mapping.html#placing-limits-on-mapped-tasks

project-defiant (approved):
I think it is a good setup for me to use it during the gwas_catalog etl step (also for other ones that require finemapping).

src/ot_orchestration/dags/ukb_ppp_finemapping.py
tskir: Indeed, but just to clarify, in this PR the _partial/expand_ routine iterates not on individual loci (of which there are potentially 100,000s in the worst case), but on chunks of the manifest, of which there are <10 in either case

PR #13, opened 2024-08-20
feat: finngen ingestion

project-defiant:
This PR implements couple of concepts:
- [x] finngen dags were revisited and updated for the r11 finngen data release - also they were deprecated from gentropy.
- [x] dynamic Batch job dag (this dag will be reused for different sources for genetic evidence generation like processing of GWAS Catalog, possibly UKBiobank, etc).
- [x]  Utility functions to write and read manifests and natively resolve paths (from previous closed PR)
- [x] Manifest operators to use these functions in the dynamic DAG.


project-defiant:
@javfg I tried to limit the changes from previous PR, this should be more digestable (still very large codebase to dig into). 
The main idea is to be able to show a runnable `genetics_dag` that will be a base for all processing of genetic evidence.

Also some changes and deprecations of existing Finngen dags, as I was working on this last week from gentropy side and updated it for new finngen r11 release.

config/config.yaml
```diff
@@ -1,39 +1,34 @@
...
+mode: CONTINUE
+gcp:
+  GCP_PROJECT: "open-targets-genetics-dev"
+  GCP_REGION: "europe-west1"
+  GCP_ZONE: "europe-west1-d"
+  GCP_DATAPROC_IMAGE: "2.1"
+  GCP_AUTOSCALING_POLICY: "otg-etl"
+
+steps:
+
```
javfg: Stray newline?

src/ot_orchestration/dags/finngen_finemapping_ingestion.py
```diff
@@ -0,0 +1,71 @@
...
+from ot_orchestration.utils import common
+from airflow.models.dag import DAG
+from airflow.utils.trigger_rule import TriggerRule
+from ot_orchestration.utils.dataproc import (
+    delete_cluster,
+    submit_step,
+    create_cluster,
+    install_dependencies,
+)
+from airflow.models.baseoperator import chain
```
javfg:
```suggestion
from pathlib import Path

from airflow.models.baseoperator import chain
from airflow.models.dag import DAG
from airflow.utils.trigger_rule import TriggerRule

from ot_orchestration.utils import common
from ot_orchestration.utils.dataproc import (
    create_cluster,
    delete_cluster,
    install_dependencies,
    submit_step,
)
```

Sort imports similarly to the rest of the files.

src/ot_orchestration/dags/genetics_dag.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import annotations
+
+from ot_orchestration.utils import read_yaml_config
+from ot_orchestration.utils.common import shared_dag_args, shared_dag_kwargs
+from airflow.models.dag import DAG
+from pathlib import Path
+
+import logging
+from ot_orchestration.dynamic_dags import generic_genetics_dag
```
javfg:
```suggestion
import logging
from pathlib import Path

from airflow.models.dag import DAG

from ot_orchestration.dynamic_dags import generic_genetics_dag
from ot_orchestration.utils import read_yaml_config
from ot_orchestration.utils.common import shared_dag_args, shared_dag_kwargs
```

Same here.

src/ot_orchestration/dags/genetics_dag.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+config_path = "/opt/airflow/config/config.yaml"
```
javfg: I know we run this with Docker, but should the full path be hardcoded here instead of being relative? If so, maybe define it as a constant in screaming snake case (love that name :D).

src/ot_orchestration/dynamic_dags.py
```diff
@@ -0,0 +1,155 @@
...
+
+import time
+from airflow.decorators import task
+import logging
+from airflow.models.taskinstance import TaskInstance
+from airflow.operators.python import get_current_context
+from airflow.exceptions import AirflowSkipException
+from airflow.utils.trigger_rule import TriggerRule
+from ot_orchestration.types import Manifest_Object
+from ot_orchestration.utils import IOManager
```
javfg:
```suggestion
import logging
import time

from airflow.decorators import task
from airflow.exceptions import AirflowSkipException
from airflow.models.baseoperator import chain
from airflow.models.taskinstance import TaskInstance
from airflow.operators.python import get_current_context
from airflow.utils.trigger_rule import TriggerRule

from ot_orchestration.operators.manifest_operators import (
    ManifestFilterOperator,
    ManifestGenerateOperator,
    ManifestReadOperator,
    ManifestSaveOperator,
    ManifestSubmitBatchJobOperator,
)
from ot_orchestration.types import Manifest_Object
from ot_orchestration.utils import IOManager
```

src/ot_orchestration/dynamic_dags.py
```diff
@@ -0,0 +1,155 @@
...
+    logging.info(params)
+    if params["mode"] == "CONTINUE":
+        return "generate_manifests"
+    else:
+        return "read_existing_manifests"
+
+
+@task(task_id="collect_task_outcome", multiple_outputs=True)
+def collect_task_outcome(manifests: list[Manifest_Object]):
+    """Collect the task(s) outcome and return failed and succeded manifests."""
```
javfg: succeeded is the right spelling

src/ot_orchestration/dynamic_dags.py
```diff
@@ -0,0 +1,155 @@
...
+        # we assume that the initial state is success
+        succeded = True
+        for key, val in new_manifest.items():
+            if key.startswith(status_flag_prefix) and not val:
+                succeded = False
+
+        if succeded:
+            succeded_manifests.append(new_manifest)
+        else:
+            failed_manifests.append(new_manifest)
```
javfg:
```suggestion
    failed_manifests = []
    succeded_manifests = []
    for new_manifest in new_manifests:
        for key, val in new_manifest.items():
            if key.startswith("pass") and not val:
                failed_manifests.append(new_manifest)
                break
        else:
            succeded_manifests.append(new_manifest)
```

What do you think about this? It should produce the same result and it stops as soon as a failed item is found. Not sure how long those lists are tho.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        studies_with_manifests: dict[str, str] = results["manifest"]["samplesheet"]
+        new_study_ids = set(studies_with_sumstats.keys()) - set(
+            studies_with_manifests.keys()
+        )
+        common_path: str = results["manifest"]["common_path"]
+        # generate the manifest per each new sumstat
+        manifests = []
+        for study_id, raw_sumstat_path in studies_with_sumstats.items():
+            if study_id in new_study_ids:
+                parital_manifest = {
```
javfg: partial_manifests

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+
+            if protocol != "gs":
+                raise NotImplementedError(
+                    "Listing objects from path with %s protocol is not implemented",
+                    protocol,
+                )
+            logging.info(
+                "Listing files at %s/%s with match glob %s", root, prefix, matchglob
+            )
+            files = GCSHook(gcp_conn_id=self.gcp_conn_id).list(
```
javfg: You should probably extract this hook into an instance attribute, or even a cached property if the execute method runs many times (see [here](https://github.com/opentargets/orchestration/blob/pis-stage/src/ot_orchestration/operators/gce.py#L606)). The way it is right now it will open a new connection every time the loop runs, and that is a pretty heavy operation (auth, etc).

src/ot_orchestration/types.py
```diff
@@ -1,21 +1,137 @@
...
+        "spark_uri": Literal["yarn"],
+        "write_mode": Literal["errorifexists", "overwrite"],
+    },
+)
+
+
+Log_Level = Literal["INFO", "TRACE", "DEBUG", "WARNING", "ERROR"]
+
+
+__all__ = [
```
javfg: You are missing `Batch_Policy_Specs`, which is being imported by utils/batch.py. In any case, I'm not sure we need to redefine `__all__` if you export everything in the file.

src/ot_orchestration/types.py
```diff
@@ -1,21 +1,137 @@
...
+    "Batch_Task_Specs",
+    "Batch_Resource_Specs",
+    "Dataproc_Specs",
+    "Batch_Specs",
+    "Provider_Name",
+    "Step_Name",
+    "DagConfigNotFound",
+    "Log_Level",
+    "Manifest_Preparation_Params",
+    "Spark_Options",
```
javfg: Is this naming mixing camelcase and underscores correct?

src/ot_orchestration/utils/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    "IOManager",
+    "GCSPath",
+    "NativePath",
+    "check_gcp_folder_exists",
+    "read_yaml_config",
+    "time_to_seconds",
+    "create_task_spec",
+    "create_batch_job",
+    "extract_study_id_from_path",
+    "generic_genetics_dag",
```
javfg: `generic_genetics_dag` has been moved to the parent directory and is missing in this module. We should consider avoiding the definition of `__all__` unless there are very good reasons because of this, it introduces a dependency you have to update whenever you change anything, and it's easy to forget.

src/ot_orchestration/utils/common.py
```diff
@@ -4,6 +4,7 @@
 
 import pendulum
 
+
```
javfg: stray newline
```suggestion
```

src/ot_orchestration/utils/batch.py
```diff
@@ -26,99 +36,113 @@ def create_container_runnable(
...
+    resources = ComputeResource()
+    resources.cpu_milli = resource_specs["cpu_milli"]
+    resources.memory_mib = resource_specs["memory_mib"]
+    resources.boot_disk_mib = resource_specs["boot_disk_mib"]
+    task.compute_resource = resources
+
+    task.max_retry_count = task_specs["max_retry_count"]
+    task.max_run_duration = Duration(
+        seconds=time_to_seconds(task_specs["max_run_duration"])
+    )
```
javfg:
```suggestion
    task.max_run_duration = f"{time_to_seconds(task_specs['max_run_duration'])}s"
```

I think you can get away with a string in here (or even return the string with the `s` straight away from `time_to_seconds`). Not sure how good it is to import the underlying protobuffer types from google cloud. This is also simpler.

But I did not check if it works, it might complain.

src/ot_orchestration/utils/batch.py
```diff
@@ -26,99 +36,113 @@ def create_container_runnable(
...
+    job = Job()
     job.task_groups = [group]
     job.allocation_policy = allocation_policy
-    job.logs_policy = batch_v1.LogsPolicy()
-    job.logs_policy.destination = batch_v1.LogsPolicy.Destination.CLOUD_LOGGING
+    job.logs_policy = LogsPolicy()
+    job.logs_policy.destination = LogsPolicy.Destination.CLOUD_LOGGING
 
     return job
+
```
javfg: Consider rewriting this as:

```
    return Job(
        task_groups=[
            TaskGroup(
                task_spec=task,
                task_environments=task_env,
            )
        ],
        allocation_policy=AllocationPolicy(
            instances=[
                AllocationPolicy.InstancePolicyOrTemplate(
                    policy=AllocationPolicy.InstancePolicy(
                        machine_type=policy_specs["machine_type"],
                        provisioning_model=AllocationPolicy.ProvisioningModel.SPOT,
                    )
                )
            ]
        ),
        logs_policy=LogsPolicy(
            destination=LogsPolicy.Destination.CLOUD_LOGGING,
        ),
    )
```

It is more straightforward to read the implicit definition than line by line, as you have to jump around to figure where things go.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+
+        Raises:
+            ValueError: when incorrect glob is defined
+            ValueError: when the glob protocol is not gs
+
+        Returns:
+            dict[str, str]: list of manifests
+        """
+        # this regex pattern can be utilized for any path or uri glob pattern
+        pattern = r"^((?P<protocol>.*)://)?(?P<root>[(\w)-]+)/(?P<prefix>([(\w)-/])+?)/(?P<matchglob>[(\w)-*]+.*){1}"
```
javfg: This is used in a couple of places, better to extract it as a constant.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+                )
+            logging.info(
+                "Listing files at %s/%s with match glob %s", root, prefix, matchglob
+            )
+            files = GCSHook(gcp_conn_id=self.gcp_conn_id).list(
+                bucket_name=root,
+                prefix=prefix,
+                match_glob=matchglob,
+            )
+            logging.info("Found %s %s files", len(files), key)
```
javfg:
```suggestion
            self.log.info("Found %s %s files", len(files), key)
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+            logging.info(files)
```
javfg:
```suggestion
            self.log.info(files)
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        super().__init__(**kwargs)
+        self.manifest_blobs = manifest_blobs
+
+    def execute(self, **_) -> list[Manifest_Object]:
+        """Execute the operator.
+
+        Returns:
+            list[Manifest_Object]: saved manifests
+        """
+        logging.info(self.manifest_blobs)
```
javfg:
```suggestion
        self.log.info(self.manifest_blobs)
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        """Read manifests.
+
+        Raises:
+            ValueError: when incorrect glob is defined.
+            NotImplementedError: for protocol other then gs.
+
+        Returns:
+            list[Manifest_Object]: list of read manifests.
+        """
+        logging.info(self.staging_manifest_path_pattern)
```
javfg:
```suggestion
        self.log.info(self.staging_manifest_path_pattern)
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        resource_specs = google_batch_specs.get("resource_specs")
+        task_specs = google_batch_specs.get("task_specs")
+        image = google_batch_specs.get("image")
+        commands = google_batch_specs.get("commands")
+        task_spec = create_task_spec(image, commands, resource_specs, task_specs)
+        task_env = [
+            Environment(variables={"MANIFEST_PATH": mp}) for mp in manifest_paths
+        ]
+        batch_job = create_batch_job(task_spec, task_env, policy_specs)
+        logging.info(batch_job)
```
javfg:
```suggestion
        self.log.info(batch_job)
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def execute(self, **_) -> list[Manifest_Object]:
+        """Execute the operator."""
+        step_flag_prefix = "pass"
+        filtered_manifests = []
+        for manifest in self.manifests:
+            for key, val in manifest.items():
+                if key.startswith(step_flag_prefix) and not val:
+                    filtered_manifests.append(manifest)
+        logging.info("PREVIOUSLY FAILED TASKS: %s", len(filtered_manifests))
```
javfg:
```suggestion
        self.log.info("PREVIOUSLY FAILED TASKS: %s", len(filtered_manifests))
```

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+from airflow.providers.google.cloud.hooks.gcs import GCSHook
+from airflow.providers.google.cloud.operators.cloud_batch import (
+    CloudBatchSubmitJobOperator,
+)
+from google.cloud.batch_v1 import Environment
+import re
+from ot_orchestration.utils.manifest import extract_study_id_from_path
+from ot_orchestration.utils.batch import create_batch_job, create_task_spec
+from ot_orchestration.types import Manifest_Object
+import logging
```
javfg:
```suggestion
```

See below (line 84)

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+
+        if protocol != "gs":
+            raise NotImplementedError(
+                "Listing objects from path with %s protocol is not implemented",
+                protocol,
+            )
+        logging.info(
+            "Listing files at %s/%s with match glob %s", root, prefix, matchglob
+        )
+        manifest_paths = GCSHook(gcp_conn_id=self.gcp_conn_id).list(
```
javfg: Here, you should use the instance method or cached hook you implement for line 84.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+            matchglob = _match.group("matchglob")
+
+            if protocol != "gs":
+                raise NotImplementedError(
+                    "Listing objects from path with %s protocol is not implemented",
+                    protocol,
+                )
+            logging.info(
+                "Listing files at %s/%s with match glob %s", root, prefix, matchglob
+            )
```
javfg:
```suggestion
            self.log.info(
                "Listing files at %s/%s with match glob %s", root, prefix, matchglob
            )
```

You can use `self.log` in Operators subclassing `BaseOperator`, as it initializes its own logger inside with Airflow custom settings.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        protocol = _match.group("protocol")
+        root = _match.group("root")
+        prefix = _match.group("prefix")
+        matchglob = _match.group("matchglob")
+
+        if protocol != "gs":
+            raise NotImplementedError(
+                "Listing objects from path with %s protocol is not implemented",
+                protocol,
+            )
```
javfg: Having an `IOManager` class which handles the loading/saving/etc from different paths, including GCS, maybe we should consider if this part belongs inside of it too.

src/ot_orchestration/operators/manifest_operators.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        # in case no manifest files, do not create the batch job
+        manifest_paths = [m["manifestPath"] for m in self.manifests]
+
+        if not manifest_paths:
+            raise AirflowSkipException("No manifests to run the batch job")
+        params = context.get("params")
+        gcp = params.get("gcp")
+        gcp_project = gcp.get("GCP_PROJECT")
+        gcp_region = gcp.get("GCP_REGION")
```
javfg: This is fetching the config from user defined params which are being populated with (I think) defaults from the yaml config file at `/opt/airflow/config/config.yaml`, but we should keep in mind we are using defaults from `utils/common.py` for these in a lot of places.

Not sure how to choose, or even if there should be a single source for these, what do you think?

src/ot_orchestration/utils/path.py
```diff
@@ -0,0 +1,345 @@
...
+class GCSPath(ProtoPath):
+    """Create GCS Path object.
+
+    Args:
+        gcs_path (str): Google Cloud Storage path.
+        chunk_size (int, optional): Chunk size for reading and writing. Defaults to 1024*256.
+        client (storage.Client, optional): Google Cloud Storage client. Defaults to storage.Client().
+    """
+
+    def __init__(self, gcs_path: str, chunk_size: int = 1024 * 256, client=None):
```
javfg: Probably better to extract `chunk_size` to a constant.

src/ot_orchestration/utils/path.py
```diff
@@ -0,0 +1,345 @@
...
+
+        Args:
+            paths (list[str]): Paths to resolve.
+
+        Returns:
+            list[ProtoPath]: List of path objects following ProtoPath protocol.
+        """
+        return [self.resolve(p) for p in paths]
+
+    def load_many(self, paths: list[str], n_threads=100) -> list[Any]:
```
javfg: Are we sure 100 threads aren't too many?

src/ot_orchestration/utils/path.py
```diff
@@ -0,0 +1,345 @@
...
+        if not paths:
+            logging.warning("Nothing to dump.")
+            return None
+
+        resolved_paths = self.resolve_paths(paths)
+        logging.info(f"DUMPING {len(paths)} OBJECTS.")
+        with concurrent.futures.ThreadPoolExecutor(max_workers=n_threads) as executor:
+            futures: list[concurrent.futures.Future] = []
+            for path, ob in zip(resolved_paths, objects, strict=True):
+                method = lambda: path.dump(ob)
```
javfg, line 420: `open()` is not thread-safe, so we must be sure no two threads are going to write to the same file. I'm guessing that should not happen, right? :)

Also, if you try to write to the same GCS file in many concurrent operations, you will get quota limited really fast (happened to me when trying to upload the same config in every step of PIS at the same time).

If we're relatively sure it won't happen, maybe we should point out this is not thread-safe in the docstring at least, so future uses keep it in mind.

javfg (changes requested):

config/config.yaml
project-defiant: true :) we need prettier for yaml files.

project-defiant: I have added yamlfmt for the pre-commit config, fortunately it will resolve this kind of issues.

src/ot_orchestration/dags/finngen_finemapping_ingestion.py
project-defiant: i did not realise, that we were missing the isort in ruff. Added it. Thanks for the point!

src/ot_orchestration/dags/genetics_dag.py
project-defiant: well, I wanted it to be external, but in the current setup it can as well sit in the package as the PIS config. I will definately change the name though

src/ot_orchestration/dynamic_dags.py
project-defiant: You are correct, my code would generate many duplicates. 😥  I must admit I will have to write tests for manifest handling.

src/ot_orchestration/types.py
project-defiant: I agree, I dropped `__all__`

src/ot_orchestration/operators/manifest_operators.py
project-defiant: It runs just twice (as many times as there are globs defined) - they are defined just for `raw files` and `manifests`. I will extract it jic

src/ot_orchestration/types.py
project-defiant: I wanted to distinguish type definitions from Failure return values https://returns.readthedocs.io/en/latest/pages/methods.html.

But this functionality was dropped. I agree we should be consistant with the naming convention. It's something I used to do. I will try to limit myself in this context.

project-defiant: btw I dropped most of these types, as I dropped the config parser, that was using it due to overcomplexifying parsing.

src/ot_orchestration/utils/__init__.py
project-defiant: I disagree, as `__all__` in the `__init__.py` increases understanding what should be considered as public. We may get away without defining it in submodules though. Also prevents from overloading when used `*` in imports as well as let's you bring the objects of functions deeply nested (this is something I cas see is missed  quiet often in Open Targets soft)

src/ot_orchestration/utils/batch.py
project-defiant: you are probably right - 
```
       max_run_duration (google.protobuf.duration_pb2.Duration):
            Maximum duration the task should run before being
            automatically retried (if enabled) or automatically failed.
            Format the value of this field as a time limit in seconds
            followed by ``s``—for example, ``3600s`` for 1 hour. The
            field accepts any value between 0 and the maximum listed for
            the ``Duration`` field type at
            https://protobuf.dev/reference/protobuf/google.protobuf/#duration;
            however, the actual maximum run time for a job will be
            limited to the maximum run time for a job listed at
            https://cloud.google.com/batch/quotas#max-job-duration.
```
but this is so not obvious and offends type definition in that field, which is sad, as I would expect that the `protobuf.MESSAGE` will map to a string type.

src/ot_orchestration/utils/batch.py
project-defiant: done

src/ot_orchestration/operators/manifest_operators.py
project-defiant: agree

src/ot_orchestration/operators/manifest_operators.py
project-defiant: Listing objects in the GCS in this context requires wildcard, which is not allowed by bulk`load` and `dump` (it requires explicit list of files) and supported indirectly (not intentionally by `load` method). 

Although I can reuse the IOManager.resolve here to resolve the pattern.

src/ot_orchestration/operators/manifest_operators.py
project-defiant: In case of using config files it is fairly easy to plug them in. I imagine that this will not change so much for our platform, it may be required if someone wants to reuse it ... in far future. Lets keep one place in the library for now (will decrease the complexity). We can always extract it to the config if needed by the community.

Edit: or when we require to run generic batch job with manifests for some other procject.

src/ot_orchestration/utils/path.py
project-defiant: That is a good question. I have followed [this](https://stackoverflow.com/questions/481970/how-many-threads-is-too-many.) and set up the default after some tests on ~40k files which took coulple of minutes. -  I was planning to make more tests when it comes to the processing of the real data.

src/ot_orchestration/utils/path.py
project-defiant: You are correct, it will not happen. I assume I can make a guard here to make sure we are not operating on the same files :)

src/ot_orchestration/utils/path.py
javfg: I'd say as long as it's not overly large and somehow slows down the machine, it's fine to approximate and then test as you said.

Ideally thread count should be similar to core count, as 1 thread per core gives the best performance (no overhead from switching context). But here we're dealing with heavy I/O locking ops, so that won't be the case.

project-defiant: I agree, I will make a branching  for it (as testing setup will not require as many threads), after all you are right that M1 has no hypertheading :p, non the less even with it it would still be of max 16 threads.

project-defiant:
@javfg ready for round 2. 🥊

javfg (approved):

project-defiant:
@javfg can you reproduce the poetry.lock issue ?

javfg:
Is it possible it's running poetry 1.8.0 in the pre-commit hook and our lockfile is 1.8.3? See [here](https://github.com/opentargets/orchestration/blob/dev/.pre-commit-config.yaml#L97) and the error we're getting in the checks.

PR #14, opened 2024-08-30
feat: pis stage for the platform orchestration dag

javfg:
This PR adds the first stage for the Platform DAG: The one that runs PIS.

In the end it is using Compute Engine machines that are spawned with [COOS](https://cloud.google.com/container-optimized-os/docs). They are very fast to spin up (~10 seconds), but quite limited on what they can do.

I've implemented an Airflow Sensor/Operator that is in charge of:

- Spawning a machine
- Attaching disks to it
- Copying config files from a bucket
- Running a docker image with some arguments
- Waiting for the exit code of that docker container
- Repoting a success (exit code 0), or a failure for the task (>0)

Most of the logic is in the gce.py file, which contains that operator.

I also:
- Added `airflow-triggerer` service to the docker-compose, as we need it for sensors.
- Added an `ignore_init` setting to interrogate so we don't have to add docstrings to `__init__` methods (it's annoying and rarely useful).

I'm opening the PR already although work is well underway for the ETL stage, as this way we can split the code into more manageable chunks for review. Shorter reviews find more things to fix!

I'll leave it as pending for now, until the merge of https://github.com/opentargets/orchestration/pull/13, so I can rebase onto it and fix any problems that arise.


javfg:
The rationale behind [remove interrogate and pydoclint](https://github.com/opentargets/orchestration/pull/14/commits/ded00c0e74a122f3cba1ab913abfa9650e7a387b) is that we are already checking docstrings with Ruff by using the [pydocstyle rules](https://docs.astral.sh/ruff/rules/#pydocstyle-d).

Interrogate and Pydoclint are extremely strict, I copypasted a docstring straight [from google code](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/compute/client_library/snippets/operations/wait_for_extended_operation.py#L35) and they complain about it. Also, `--ignore-init-function` and `--style google` are mutually exclusive and it is a pain to document all `__init__` functions.

Let me know if you think this is too big a change.

javfg:
I cannot reproduce the check error on the `poetry lock` pre-commit check. We should discuss on loosening those, right now it seems a bit overkill. :)

.pre-commit-config.yaml
```diff
@@ -67,29 +67,17 @@ repos:
...
         args: [--all]
   - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
-    rev: v9.16.0
+    rev: v9.17.0
     hooks:
       - id: commitlint
-        additional_dependencies: ["@commitlint/config-conventional@18.6.3"]
+        additional_dependencies: ['@commitlint/config-conventional@18.6.3']
         stages: [commit-msg]
-  # Check for docstring coverage in python files.
```
project-defiant, line 75: ruff performs  mostly the same checks, I agree we can drop it. That is what I was aiming for in the future, to swap these two by ruff lint

docker-compose.yaml
```diff
@@ -99,7 +99,7 @@ services:
       start_period: 5s
     restart: always
   airflow-webserver:
-    !!merge <<: *airflow-common
```
project-defiant, line 102: very curious why it was added by yamlfmt and now removed. I assume we can skip the !!merge, as the result should be the same.

docker-compose.yaml
```diff
@@ -192,16 +192,32 @@ services:
...
-      !!merge <<: *airflow-common-env
+      <<: *airflow-common-env
       CONNECTION_CHECK_MAX_COUNT: "0"
     # Workaround for entrypoint issue. See: https://github.com/apache/airflow/issues/16252
     command:
       - bash
       - -c
       - airflow
+
+  airflow-triggerer:
```
project-defiant, line 207: Can you point to the docs? I spend some time and found nothing on the airflow page.

src/ot_orchestration/dags/platform_input_support.py
```diff
@@ -1,96 +1,102 @@
...
+            description=(
+                "The tag of the platform-input-support image to use. Defaults to 'latest', "
+                "but it is recommended to use a specific tag. Keep in mind tags do not have the 'v' prefix."
+            ),
+        ),
+    },
 ) as dag:
-    for step_name in get_steps():
-        clean_step_name = clean_name(step_name)
+    for step in list(CONFIG["steps"].keys()):
```
project-defiant, line 51: why not pass `CONFIG`. as a default param to the `DAG`?

src/ot_orchestration/dags/platform_input_support.py
```diff
@@ -1,96 +1,102 @@
...
+
+        @task_group(group_id=f"pis_{step}")
+        def pis_step(step: str) -> None:
+            instance_name = f"uo-pis-{clean_name(step)}-{{{{ run_id | strhash }}}}"
```
project-defiant, line 55: nice to know that you can substitute any Jinja template in airflow templates.

src/ot_orchestration/dags/platform_input_support.py
```diff
@@ -1,96 +1,102 @@
...
-                d = CloudRunDeleteJobOperator(
-                    task_id=f"{task_instance.task_id}_delete",
-                    project_id=PIS_GCP_PROJECT,
-                    region=GCP_REGION,
-                    job_name=job_name,
-                    trigger_rule="all_done",
-                    dag=dag,
-                )
-                d.execute(context=task_instance.get_template_context())
+            c >> Label("invalid previous run") >> u >> r >> d >> j
```
project-defiant, line 99: Can you name the tasks?

src/ot_orchestration/dags/platform_input_support.py
```diff
@@ -1,96 +1,102 @@
...
+                container_image=f"{PIS_IMAGE}:{dag.params['pis_image_tag']}",
+                container_env={
+                    "PIS_STEP": step,
+                    "PIS_CONFIG_FILE": "/" + config_filename,
+                    "PIS_POOL": "16",
+                },
+                container_service_account=PIS_SACC,
+                container_scopes=["https://www.googleapis.com/auth/drive"],
+                container_files={f"{GCS_URL}/{config_filename}": "/" + config_filename},
+                work_disk_size_gb=150,
```
project-defiant, line 84: why size of 150gb?

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+            try:
+                entries = super().list_entries(*args, **kwargs)
+                break
+            except ResourceExhausted:
+                self.log.warning(
+                    "Rate limit for logging api exceeded, waiting for %d seconds",
+                    self.request_interval,
+                )
+                time.sleep(self.request_interval)
+                self.request_interval *= 2
```
project-defiant, line 120: This time command could explode pretty fast. Mayby add another `LOGGING_REQUEST_INTERVAL`, this way it will be linear, not exponential grow of the request interval.

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+
+        return entries
+
+
+class CloudLoggingHook(GoogleBaseHook):
+    """Hook for the Google Logging service.
+
+    :param gcp_conn_id: The connection ID to use when fetching connection info.
```
project-defiant: Let's stick to a single docstring style - I have found that one can use [pyment](https://github.com/dadadel/pyment/blob/master/doc/sphinx/source/pyment.rst#customization) to convert from one to another.

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+                credentials=self.get_credentials(),
+                client_info=CLIENT_INFO,
+            )
+        return self._client
+
+
+class CloudLoggingAsyncHook(GoogleBaseHook):
+    """Async hook for the Google Logging service.
+
+    :param gcp_conn_id: The connection ID to use when fetching connection info.
```
project-defiant: Please change to the same docstring style.

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+        log_pages = None
+
+        while True:
+            try:
+                log_pages = await client.list_log_entries(
+                    resource_names=[f"projects/{project_name}"],
+                    filter=query,
+                    timeout=300,
+                )
+                break
```
project-defiant, line 243: I am not sure, but sleeping inside the while loop makes a bit more sense to me. It may be less optimised, but would allow us to have more granular control on the number of Exceptions.

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+    uses a startup script to run the container passed in `container_image`. Arguments and environment
+    can be passed to the container using `container_args` and `container_env` respectively. The
+    sensor will wait for the startup script to finish and return the exit code.
+
+    Be aware this sensor _MUST_ run in deferrable mode (explicitly setting `deferrable=True`). The
+    poke method is not implemented, and the sensor will never return True when run in blocking mode.
+
+    To enable non-deferrable mode we must implement the poke method properly.
+
+    :param project_id: Optional, the Google Cloud project ID where the job is.
```
project-defiant: docstring format

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+    :param container_args: Arguments to pass to the container (optional).
+    :param container_env: Environment variables to pass to the container (optional).
+    :param container_service_account: Service account to use for the instance (default default).
+    :param container_scopes: A list of extra scopes to add to the service account if any are needed.
+    :param container_files: Files to copy to the instance (optional). This is a dictionary where
+        the key is a GCS path in the form `gs://bucket/path/to/file` and the value is the path
+        where the file will be copied to in the instance. This is intented for small files needed
+        to run the workload (like the configuration). Large files should be downloaded by the
+        workload itself. The paths specified as values will be relative to the `/home/app` directory
+        and will be created if they don't exist. Inside the docker container, they will be mounted
```
project-defiant:
```suggestion
        and the parent paths will be created if they don't exist. Inside the docker container, they will be mounted
```

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+    :param gcp_conn_id: The connection ID to use connecting to Google Cloud.
+    :param impersonation_chain: Optional service account to impersonate using short-term
+        credentials, or chained list of accounts required to get the access_token
+        of the last account in the list, which will be impersonated in the request.
+        If set as a string, the account must grant the originating account
+        the Service Account Token Creator IAM role.
+        If set as a sequence, the identities from the list must grant
+        Service Account Token Creator IAM role to the directly preceding identity, with first
+        account from the list granting this role to the originating account (templated).
+    :param deferrable: If True, run the sensor in deferrable mode.
```
project-defiant: If it's only implemented for defferable mode, raise NotImplementedError when defferable is set to False

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+            dest=( {dest_paths} )
+            if [ -n "${{dest[*]}}" ]; then
+                for i in "${{!dest[@]}}"; do
+                    if=${{orig[$i]}}
+                    od=$(dirname "${{dest[$i]}}")
+                    if [ "$od" = "." ]; then
+                        od=""
+                    else
+                        mkdir -p "$od"
+                    fi
```
project-defiant: I think you can just get away with 
```suggestion
                   mkdir -p "$od"
```

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+            {init_work_disk}
+            useradd -m app
+            usermod -a -G docker app
+            mkdir -p /home/app
+            cd /home/app
+            orig=( {gcs_files} )
+            dest=( {dest_paths} )
+            if [ -n "${{dest[*]}}" ]; then
+                for i in "${{!dest[@]}}"; do
+                    if=${{orig[$i]}}
```
project-defiant: please change the var name 😨

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+                    od=$(dirname "${{dest[$i]}}")
+                    if [ "$od" = "." ]; then
+                        od=""
+                    else
+                        mkdir -p "$od"
+                    fi
+                    of=${{dest[$i]}}
+                    echo "Copying $if to $of ($od)"
```
project-defiant, line 436: The VM container itself does not have the `gsutil` I assume. Nice inception

src/ot_orchestration/operators/gce.py
```diff
@@ -0,0 +1,733 @@
...
+        return CloudLoggingHook(
+            gcp_conn_id=self.gcp_conn_id,
+            impersonation_chain=self.impersonation_chain,
+        )
+
+
+class ComputeEngineExitCodeTrigger(BaseTrigger):
+    """Trigger that checks for the exit code of a google compute engine instance startup script.
+
+    :param instance_name: Name of the instance to check.
```
project-defiant: docstring change

src/ot_orchestration/operators/platform.py
```diff
@@ -0,0 +1,103 @@
...
+    that are relevant to the current step, comparing them against the copy in the
+    specified bucket. If they are different, that means the step should be
+    run.
+
+    It would be interesting add a check that downloads the manifest from the
+    specified bucket and compares the resources listed for the step with the
+    files in the bucket, along with their checksums. That way we can ensure the
+    files exist and have not been tampered with.
+
+    :param project_id: The GCP project ID. Defaults to the platform project.
```
project-defiant: docstring change

src/ot_orchestration/operators/platform.py
```diff
@@ -0,0 +1,103 @@
...
+
+        try:
+            upstream_config = self.get_upstream_config(self.upstream_config_url)
+        except NotFound:
+            self.log.info(
+                "Upstream configuration file not found, assuming first run, %s",
+                self.upstream_config_url,
+            )
+
+        lc = self.extract_relevant_config(local_config)
```
project-defiant, line 92: The step parameters change will not trigger the rerun in this context.

src/ot_orchestration/utils/utils.py
```diff
@@ -10,6 +11,16 @@
 from google.cloud.storage import Client
 
 
+def bucket_name(gs_url: str) -> str:
```
project-defiant: This could be used in `GCSPath` (in case you require the full path object, I you might want to add a method there.

yamlfmt.yml
```diff
@@ -1,3 +1,4 @@
 formatter:
   type: basic
   retain_line_breaks_single: true
+  drop_merge_tag: true
```
project-defiant, line 4: I see why `!!merge` flags were skipped now

project-defiant (commented):
Some minor comments, otherwise well done!

docker-compose.yaml
javfg: This is to be able to run deferrable triggers, which asynchronously check something and are offloaded from the worker between checks.

You can check the requirement [here](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/deferring.html#using-deferrable-operators) and the [official docker compose](https://github.com/apache/airflow/blob/main/docs/apache-airflow/howto/docker-compose/docker-compose.yaml#L174).

src/ot_orchestration/dags/platform_input_support.py
javfg: I thought about it, but the config can change between executions (not runs, but partial reruns of some failed steps). If I put the config in a param it will be immutable for the whole run, and then you'd have to rerun the whole DAG if something requiring a config change fails.

docker-compose.yaml
project-defiant: Thught that healthcheck was somehow derived from airflow, I did not know about it in the compose. Nice!

src/ot_orchestration/dags/platform_input_support.py
javfg: It's what's required for the biggest size step (openfda). In the next PR I've refactored this and extracted that into an easier to find constant.

src/ot_orchestration/operators/gce.py
javfg: I did this on purpose, because if the request fails for a quota limit, it might be necessary to wait up to a minute. We perform less requests if exponentially increasing than if we check every x seconds.

src/ot_orchestration/operators/gce.py
javfg: Keep in mind the try/catch is there mostly because of the cloud logging quota limit, which is 60 requests per minute per service account. If I sleep after every request, instead of taking the amount of time per request (usually milliseconds) * number of requests, this will take the amount of time per request + the interval * number of requests. That can be a whole lot more. :)

My idea was to start sleeping only when we reach that limit, which does not happen often but once it happens will repeat often unless we stop requesting for a while (hence the exponential backoff).

src/ot_orchestration/operators/platform.py
javfg: Can you clarify this one? I don't understand what you mean.

src/ot_orchestration/dags/platform_input_support.py
javfg: As discussed, the new comments explain what those letters are in a bit more detail and we can keep them like that as they are only used right after being defined.

src/ot_orchestration/utils/utils.py
javfg: Indeed, actually we can use the `split` method you created. I've updated my code.

src/ot_orchestration/dags/platform_input_support.py
project-defiant: @javfg I wonder if then it could be referenced in the step config.

src/ot_orchestration/operators/gce.py
project-defiant: I agree, but then again you do not recover from it after successful attempt, which will prolong the next consecutive checks.

src/ot_orchestration/dags/platform_input_support.py
project-defiant: You can specify it when triggering the dag then. The values you pass to the params are only the defaults in that context.

src/ot_orchestration/operators/gce.py
project-defiant: Again, I agree, but I would still reset the interval after successful attempt.

src/ot_orchestration/operators/platform.py
project-defiant: Imagine case when you just change the step config for instance from
```
    - name: download mondo
      source: https://github.com/monarch-initiative/mondo/releases/download/v2021-12-30/mondo.json
      destination: ontology-inputs/ontology-mondo.json
```
to 
```
    - name: download mondo
      source: https://github.com/monarch-initiative/mondo/releases/download/v2024-12-12/mondo.json
      destination: ontology-inputs/ontology-mondo.json
```
(change in the date of mondo source). 

Will it trigger the rerun in that context?

src/ot_orchestration/dags/platform_input_support.py
javfg: Imagine this situation:

You run the Pipeline DAG with a config for PIS which fails for one step only. That step is left red in the DAG and you want to rerun it after editing the config file.

If you pass the config into the DAG as a param, it will be the same when we go into Airflow UI and clean the failed step so it runs again. The changes will not be picked up and it will fail again.

src/ot_orchestration/dags/platform_input_support.py
javfg: I would not, this is not going to change frequently. It should be enough to have it together with the other constants.

src/ot_orchestration/operators/gce.py
javfg: Oh, that is a very good point! I'll reset the interval when a request is successful.

src/ot_orchestration/operators/platform.py
javfg: It should, it gets both the scratchpad and the dict for the step that is running (Lines 72-75) and decides the branch based on whether any of those two dictionaries are different (line 96).

Am I missing something?

src/ot_orchestration/dags/platform_input_support.py
project-defiant: true, that it won't change frequently, but I imagine that every step do not have the same requirement of 150Gb disk usage, am I correct in this assumption?

src/ot_orchestration/operators/platform.py
project-defiant: my mistake, you are correct. I missed nested `get` when you extract the step config.

src/ot_orchestration/dags/platform_input_support.py
project-defiant: You almost convinced me. Let's keep it outside, although I want to test this behavior.

src/ot_orchestration/dags/platform_input_support.py
javfg: You're correct, but step machines (and their disks) live for average 2 to 3 minutes, to a maximum of 17 minutes for one of them. The cost difference for that time span is zero, as far as I can tell.

Although I cannot say this 100% sure. There is no trace of the disk cost for PIS in the billing even if I'm labeling the disks. This could be because they are added into the VM cost because they are attached disks.

If I'm labeling everything properly, running PIS is about £3.40.

src/ot_orchestration/dags/platform_input_support.py
javfg: You can see the values for the params for a given run in the 'details' tab when clicking on a run. Those will be fixed for the whole run and will not change if you rerun parts of the graph.

image

project-defiant (approved):

project-defiant:
@javfg approved. Nice work!

project-defiant:
@javfg can you drop the `poetry.lock` check. I had the same issue with it on my previous PR.

PR #15, opened 2024-09-09
feat: extract config for finngen ingestion to yaml file

project-defiant:
# Context

Move all configuration of the finngen_ingestion dag to the separate config.
This dag will has it's own lifecycle depending on the [finngen data releases](https://finngen.gitbook.io/documentation), so it's a good idea to have it run on demand only when requred and cache the produced results:
- finngen credible sets (ingested by gentropy StudyLocus class)
- finngen study index (ingested by gentropy StudyIndex class)
inside the staging gcs path `gs://finngen_data/` with appropriate release prefix (`r11`)

# Implementations
- [x] extracted constants from the `finngen_ingestion.py` to `finngen_ingestion.yaml` config file.
- [x] Explicitly declared parameters that were used as default in gentropy finngen steps.
- [x] Renamed dag


src/ot_orchestration/dags/finngen_ingesion.py
javfg, line 1: There is a typo in the filename. `finngen_ingesion.py` -> `finngen_ingestion.py`.

src/ot_orchestration/dags/finngen_ingesion.py
```diff
@@ -15,18 +15,10 @@
     install_dependencies,
     submit_step,
 )
+from ot_orchestration.utils.utils import read_yaml_config
```
javfg:
```suggestion
from ot_orchestration.utils import read_yaml_config
```

You export `read_yaml_config` in the `__init__` file.

javfg (changes requested):

src/ot_orchestration/dags/finngen_ingesion.py
project-defiant: you are right, I am not too consistent about my `__all__` flags :p. This was IDE that pointed be out to the wrong subpackage.

project-defiant:
@javfg ready for 🥊  2

javfg (approved):

javfg:
All good this time around!

PR #16, opened 2024-09-10
feat: Allow gcs paths without prefix

project-defiant:
## Context

I want to use google cloud storage paths that do not have prefix like `gs://gwas_catalog_data/manifests/` with the `GCSPath` object.  This will be required for data relocation within the `genetics_etl` dag.

Currently `GCSPath` object regex requires path to have `prefix` and `filename` groups which prevents us from using paths without the prefix, or paths that point directly to the directories within this class.

## Implementation

This PR tries to solve it by changing the regex pattern to catch `path` instead of `filename and prefix` and extract the logic behind path`filename` and `prefix` parsing to the `GCSPath.segments` method.

- [x] Refactored `POSIX_PATH_PATTERN` to capture `path` instead of `prefix` and `filename`.
- [x] Refactored `GCSPath.segments` and `GCSPath.path` methods to preserve backwards compatibility.
- [x] Added tests for `GCSPath` object.

### DIFF
--- a/src/ot_orchestration/utils/path.py
+++ b/src/ot_orchestration/utils/path.py
@@ -1,7 +1,5 @@
 """Module for Google Cloud Path naive parsers."""
 
-from __future__ import annotations
-
 import concurrent.futures
 import json
 import logging
@@ -18,7 +16,7 @@ from requests.adapters import HTTPAdapter
 
 CHUNK_SIZE = 1024 * 256
 MAX_N_THREADS = 32
-POSIX_PATH_PATTERN = r"^^((?P<protocol>.*)://)?(?P<root>[(\w)-]+)/(?P<path>([(\w)-/])+)"
+URI_PATTERN = r"^^((?P<protocol>.*)://)?(?P<root>[(\w)-]+)/(?P<path>([(\w)-/])+)"
 
 
 class PathSegments(TypedDict):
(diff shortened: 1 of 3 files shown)


src/ot_orchestration/utils/path.py
```diff
@@ -1,5 +1,7 @@
 """Module for Google Cloud Path naive parsers."""
 
+from __future__ import annotations
```
javfg: Are you using this for any change introduced here?

src/ot_orchestration/utils/path.py
```diff
@@ -16,7 +18,7 @@
 
 CHUNK_SIZE = 1024 * 256
 MAX_N_THREADS = 32
-POSIX_PATH_PATTERN = r"^((?P<protocol>.*)://)?(?P<root>[(\w)-]+)/(?P<prefix>([(\w)-/])+?)/(?P<filename>[(\w)-*]+.*){1}"
+POSIX_PATH_PATTERN = r"^^((?P<protocol>.*)://)?(?P<root>[(\w)-]+)/(?P<path>([(\w)-/])+)"
```
javfg: This is not a POSIX path, I'd say it's a URI or URL.

javfg (changes requested):

src/ot_orchestration/utils/path.py
project-defiant: I was wondering how to name it, as the regex would also be able to catch the path without the protocol, so pure POSIX, but the way we use it now focuses purely on URI, so I agree, lets make it `URI_PATH_PATTERN`

src/ot_orchestration/utils/path.py
project-defiant: I do not use it at all. I wonder how it got there...

javfg (approved):

javfg:
Looks good!

PR #17, opened 2024-09-10
feat: move ukb_ppp_eur datasource ingestion to orchestration

project-defiant:
## Context

We want to move all dags from `gentropy` to the `orchestration` repositoy.

## Implementations

This PR:
- [x] moves the dag to ingest and harmonize the `UKB PPP EUR` dataset. 
- [x] Creates dag config and extracts all configuration parameters to it
- [x] Adds a function to convert parameters from dag config to gentropy step parameters (handled by hydra).
- [x] Refactors dag to allow for addition of finemapping step.

### DIFF
--- a/src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
+++ b/src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
@@ -1,4 +1,4 @@
-cluster_name: otg-ukb-ppp-eur"
+cluster_name: otg-ukb-ppp-eur
 steps:
   - id: ukb_ppp_eur_sumstat_preprocess
     params:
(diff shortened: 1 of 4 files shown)


src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
```diff
@@ -0,0 +1,12 @@
+cluster_name: otg-ukb-ppp-eur"
```
tskir:
```suggestion
cluster_name: otg-ukb-ppp-eur
```

tskir (changes requested):
Everything looks good except for the one typo in `cluster_name`

tskir (approved):
Now fixed the cluster_name, should be good to merge!

PR #18, opened 2024-09-11
feat: add validation to genetics etl

project-defiant:
## Context
https://github.com/opentargets/issues/issues/3433

## Implementations
- [x] Added validation steps to the `genetics_dag`.
- [x] Streamlined the `generate_dag` function, so the `create_cluster` task is  upstream of tasks that do not have any upstream tasks (starting nodes in the DAG) and `delete_cluster` is downstream of tasks that do not have any other downstream tasks (ending nodes in the DAG)

## Missing

The next steps would be to move the individual configurations of the ETL steps to the `genetics_etl.yaml`


src/ot_orchestration/dags/genetics_etl.py
```diff
@@ -10,108 +10,37 @@
...
+        "destination_bucket": release_dir.bucket,
+        "destination_object": f"{release_dir.path}/manifests/",
     },
     # L2G gold standard:
-    "gold_standard": {
-        "source_bucket": RELEASE_BUCKET_NAME,
-        "source_object": "input/l2g/gold_standard/curation.json",
-        "destination_bucket": RELEASE_BUCKET_NAME,
-        "destination_object": f"releases/{RELEASE_VERSION}/locus_to_gene_gold_standard.json",
+    "l2g_gold_standard": {
```
ireneisdoomed, line 39: Not the purpose of this PR, but this should be moved to the curation repo. It's a static file.

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+cluster_name: otg-etl
+gwas_catalog_manifests_path: gs://gwas_catalog_data/manifests
+l2g_gold_standard_path: gs://genetics_etl_python_playground/input/l2g/gold_standard/curation.json
+release_dir: gs://genetics_etl_python_playground/releases/24.09
+
+steps:
+  - id: study_validation
+    params:
+      study_index_path:
+        - gs://gwas_catalog_data/study_index
```
ireneisdoomed, line 10: The same way we are versioning Finngen (R11), should we version the other datasources?

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+        - gs://eqtl_catalogue_data/study_index
+        - gs://ukb_ppp_eur_data/study_index
+        - gs://finngen_data/r11/study_index
+      target_index_path: gs://genetics_etl_python_playground/releases/24.06/gene_index
```
ireneisdoomed, line 14: It's very likely we will deprecate Gentropy's gene index, and stick with the target index from the Platform. Is it possible to use the one from the Platform.
Isn't there a circular dependency problem here? The ETL needs Gentropy ETL outputs, but to distinguish Gentropy valid and invalids we should use the Platform ETL outputs.

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+      disease_index_path: gs://open-targets-pre-data-releases/24.06/output/etl/parquet/diseases
```
ireneisdoomed, line 15: Same comment, this should be the latest disease index.

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+      valid_study_index_path: &valid_study_index gs://genetics_etl_python_playground/releases/24.09/study_index
```
ireneisdoomed, line 16: That's a cool keyword I didn't know about :D

ireneisdoomed:
```suggestion
      valid_study_index_path: &valid_study_index ${release_dir}/study_index
```

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+      invalid_study_index_path: gs://genetics_etl_python_playground/releases/24.09/invalid_study_index
```
ireneisdoomed, line 17:
```suggestion
      invalid_study_index_path: ${release_dir}/invalid_study_index
```

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+    params:
+      study_index_path: *valid_study_index
+      study_locus_path:
+        - gs://gwas_catalog_data/credible_set_datasets/gwas_catalog_PICSed_curated_associations
+        - gs://gwas_catalog_data/credible_set_datasets/gwas_catalog_PICSed_summary_statistics
+        - gs://gwas_catalog_data/credible_set_datasets/gwas_catalog_susie_summary_statistics
+        - gs://eqtl_catalogue_data/credible_set_datasets/eqtl_catalogue_susie/
+        - gs://ukb_ppp_eur_data/credible_set_datasets/ukb_ppp_eur_susie
+        - gs://finngen_data/r11/credible_set_datasets/finngen_r11_susie
+      valid_study_locus_path: &valid_credible_set gs://genetics_etl_python_playground/releases/24.09/credible_set
```
ireneisdoomed, line 38:
```suggestion
      valid_study_locus_path: &valid_credible_set ${release_dir}/credible_set
```

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+        - FAILED_STUDY
+        - MISSING_STUDY
+        - NO_GENOMIC_LOCATION_FLAG
+        - COMPOSITE_FLAG
+        - INCONSISTENCY_FLAG
+        - PALINDROMIC_ALLELE_FLAG
+        - SUBSIGNIFICANT_FLAG
+      session.write_mode: overwrite
+  - id: "ot_gene_index"
+  - id: "ot_variant_index"
```
ireneisdoomed, line 52: This won't work with the latest variant index

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+  - id: "ot_colocalisation_ecaviar"
+    prerequisites:
+      - credible_set_validation
+  - id: "ot_colocalisation_coloc"
+    prerequisites:
+      - credible_set_validation
+  - id: "ot_locus_to_gene_train"
+    prerequisites:
+      - credible_set_validation
+      - "ot_variant_index"
```
ireneisdoomed, line 67: Same as above

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+        - INCONSISTENCY_FLAG
+        - PALINDROMIC_ALLELE_FLAG
+        - SUBSIGNIFICANT_FLAG
+      session.write_mode: overwrite
+  - id: "ot_gene_index"
+  - id: "ot_variant_index"
+  - id: "ot_variant_to_gene"
+    prerequisites:
+      - credible_set_validation
+      - "ot_variant_index"
```
ireneisdoomed, line 56: Same as abive

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -1,21 +1,78 @@
...
+      - "ot_variant_to_gene"
+      - "ot_colocalisation_ecaviar"
+      - "ot_colocalisation_coloc"
+  - id: "ot_locus_to_gene_predict"
+    prerequisites:
+      - credible_set_validation
+      - "ot_locus_to_gene_train"
+      - "ot_variant_index"
```
ireneisdoomed, line 75: Same as above

ireneisdoomed (commented):
The DAG architecture looks fine to me. It's neat that changes are mostly configuration, and not code. 
My major question/concern is whether we have a circular dependency with the Platform ETL, more in the comments below. UPDATE: ok, clarified now that Gentropy and Platform will be intertwined. 
Another question, maybe for this one, is the compatibility between the ETL and the variant index DAG, that should be embedded in the ETL one and it uses operators other than the Dataproc job submitter.

ireneisdoomed:
Btw, could you please change the `make dev` rule? Use `.` instead of `source`. It was failing

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: As a matter effect, we do not use the versioning there for now, although if we want to update them it would be great to do that

ireneisdoomed (approved):
Thank you for the discussion

src/ot_orchestration/dags/genetics_etl.py
project-defiant: I am linking the other PR -> https://github.com/opentargets/curation/pull/36

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: As we discussed. The Genetics ETL will be cut into pieces and mixed within the Platform ETL. This way we will have no trouble when running different steps that have in-dag dependencies between processes.

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: It is very explicit now, but the main idea is to understand the inputs and outputs. If we can do that for entire dag, we then can figure out the best way to make the substitutions with placeholders. The config does not use OmegaConf for now but I get your point that we could use it eventually.

I will migrate it to use [variable interpolation](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#variable-interpolation), but only after all configs are `static` so we can have a clear picture what should be interpolated and clearly used to control the state of the DAG.

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: That is true, but I am making another PR to merge the old vi with this approach.

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: as above

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: as above

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: as  above

project-defiant:
> Btw, could you please change the `make dev` rule? Use `.` instead of `source`. It was failing
@ireneisdoomed Done!

PR #22, opened 2024-09-12
feat: add FinnGen UKB meta-analysis ingestion DAG

tskir:
Also fixed the `extended_stark_conf` section syntax for UKB PPP data

### DIFF
--- a/src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
+++ b/src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
@@ -2,11 +2,12 @@ cluster_name: otg-ukb-ppp-eur
 steps:
   - id: ukb_ppp_eur_sumstat_preprocess
     params:
-      raw_study_index_path: gs://gentropy-tmp/batch/output/ukb_ppp_eur/study_index.tsv
+      raw_study_index_path_from_tsv: gs://gentropy-tmp/batch/output/ukb_ppp_eur/study_index.tsv
       raw_summary_stats_path: gs://gentropy-tmp/batch/output/ukb_ppp_eur/summary_stats.parquet
       variant_annotation_path: gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/variant_annotation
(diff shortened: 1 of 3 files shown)


src/ot_orchestration/dags/config/ukb_ppp_eur_harmonisation.yaml
```diff
@@ -2,11 +2,12 @@ cluster_name: otg-ukb-ppp-eur
...
+      raw_study_index_path_from_tsv: gs://gentropy-tmp/batch/output/ukb_ppp_eur/study_index.tsv
       raw_summary_stats_path: gs://gentropy-tmp/batch/output/ukb_ppp_eur/summary_stats.parquet
       variant_annotation_path: gs://genetics_etl_python_playground/output/python_etl/parquet/XX.XX/variant_annotation
       tmp_variant_annotation_path: gs://gentropy-tmp/variant_annotation
       study_index_output_path: gs://ukb_ppp_eur_data/study_index
       summary_stats_output_path: gs://ukb_ppp_eur_data/summary_stats
-      # session parameters
-      session.extended_stark_conf.spark.sql.shuffle.partitions: "3200"
+session:
+  extended_stark_conf:
```
project-defiant:
```suggestion
  extended_spark_conf:
```

project-defiant: I may be wrong but the `session.extended_spark_conf.spark.sql.shuffle.partitions` with a typo was on purpose. This config is converted by the dag and passed as command line options to the gentropy step, so it does not take an advantage of hydra structured config. The fact that it was not working was possibly due to the typo in the name.

tskir: Oooh you're right! I totally missed the typo. I'll revert this part back to how it was (& will fix the typo)

tskir: Actually, I'll do one better. Given that it worked well with the typo (so the value definitely wasn't populated and processed), I'm pretty sure that this parameter isn't required for efficient processing. So I'll just remove it.

PR #24, opened 2024-09-16
feat: add ETL stage to the platform pipeline

### LINKED ISSUE #3396: Port the ETL launch to Airflow
This task is under the context of the unification of the Gentropy and Platform data pipelines (#3394).

# Description of the work
Extend the [`platform-input-support` graph](https://github.com/opentargets/orchestration/commit/124339341ec76a625b9d91b5f0ded92c8d65cca3) in Airflow with the ETL steps.

Currently, determining the order of execution of the ETL steps is handled by Google Dataproc. To be able to integrate it in the unified pipeline, we must describe this dependency graph in [the orchestrator](https://github.com/opentargets/orchestration).

This is the dependency graph for the ETL steps, including the dependencies from PIS steps:

etl-dependencies-2

# Tasks and current status
- [x] Determine the dependencies between each step in the ETL
- [x] Determine the dependencies ETL steps have on PIS steps, to be able to plug both parts
- [x] Extend the orchestrator `platform-input-support` graph with tasks for the ETL

javfg:
This PR adds the ETL stage to the platform pipeline, completing the platform side of things in the orchestrator.

There are few extra additions. I've separated the code different commits which should not be squashed when merging. The important ones are:

- d3a8a63: New [HOCON](https://docs.spongepowered.org/stable/en/server/getting-started/configuration/hocon.html) parse features. This is the config format used by the ETL application.
- f34ee7b: New `Labels` class that provides various utilities for handling Google Cloud labels.
- b7f3701: New operators to handle file upload, from a string and from a remote location using a URL.
- 79b1ee1: Dataproc operators to handle creation of clusters and submission of jobs. They simplify these tasks by having our defaults put in. These are tailored for ETL but could easily be made more generic for other parts.
- 7390670: Some changes to the way we decide if we rerun a PIS step in subsequent run tries.

The main commit is 9af6ca1, where the `platform-input-support` dag is removed and `platform` is put in its place with the full pipeline. Some details:

- The idea is to have a `platform.yaml` defining the run parameters. These are the things we change with each release, like bucket path, efo/cheml/ensembl versions, etc.
- There is a `platform.py` file in the configuration which will provide all the values the DAG needs by using the previous yaml to override values in PIS/ETL configurations. That way, we don't have to edit those for each release.

I will detail this more in a design document later one, hopefully :).

This closes https://github.com/opentargets/issues/issues/3396.


pyproject.toml
```diff
@@ -26,6 +26,7 @@ pendulum = "^3.0.0"
 apache-airflow-providers-apache-beam = "^5.7.1"
 typing-extensions = "^4.12.2"
 requests = "^2.32.3"
+pyhocon = "^0.3.61"
```
project-defiant, line 29: So we eventually need some config parser tool :)

src/ot_orchestration/dags/config/platform.py
```diff
@@ -0,0 +1,131 @@
+"""Configuration class for the platform pipeline."""
```
project-defiant, line 1: For the sake of convention to store only data files in the `dags/config`, could you move `.py` files to another subpackage ? Mayby config_parsers ?

src/ot_orchestration/dags/config/platform.py
```diff
@@ -0,0 +1,131 @@
...
+
+    The configuration files are expected to be in the same directory as this file.
+    They are:
+    - `platform.yaml`: contains the general configuration for the pipeline.
+    - `pis.yaml`: contains the configuration for the PIS steps.
+    - `etl.conf`: contains the configuration for the ETL steps.
+    """
+
+    def __init__(self) -> None:
+        self.platform_config_path = Path(__file__).parent / "platform.yaml"
```
project-defiant, line 27: In case we ever build the package, we might have to include all configs as data files.

src/ot_orchestration/dags/config/platform.py
```diff
@@ -0,0 +1,131 @@
...
+        self.service_account_scopes = ["https://www.googleapis.com/auth/drive"]
+        # fmt: on
+
+        # Platform pipeline settings.
+        settings = read_yaml_config(self.platform_config_path)
+        self.gcs_url = settings["gcs_url"]
+        self.chembl_version = settings["chembl_version"]
+        self.efo_version = settings["efo_version"]
+        self.ensembl_version = settings["ensembl_version"]
+        self.is_ppp = settings["is_ppp"]
```
project-defiant, line 55: I wonder if we should drop this flag and insead have two separate versions of the configuration living in different branches (versions).

src/ot_orchestration/dags/config/platform.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # PIS-specific settings.
+        self.pis_config = self.init_pis_config()
+        pis_version = settings["pis_version"]
+        self.pis_image = f"{pis_image_base}:{pis_version}"
+        self.pis_step_list = self.pis_config["steps"].keys()
```
project-defiant, line 63: NIT, as we will have some name clashes in the future when merging the genetics_etl config, I think we should settle down the nomenclature. I initially called the individual steps in the config the same way as you did - `steps`, but when I had to distinguish between the `step` and `task_group` (that not necesarly mapped to any of the steps introduced by the gentropy) In that context I have changed the name from `steps` to `nodes` as this name have wider representation in my opinion.

src/ot_orchestration/dags/config/platform.py
```diff
@@ -0,0 +1,131 @@
...
+        return pis_raw_conf
+
+    # pyhocon returns a ConfigTree, but we can treat it as a dict
+    def init_etl_config(self) -> Any:
+        """Initialize the ETL configuration.
+
+        This method reads the ETL configuration file, replaces the fields defined
+        in the orchestrator config, and returns the resulting configuration.
+        """
+        etl_raw_conf = read_hocon_config(self.etl_config_local_path)
```
project-defiant, line 114: I guess this is why we need hocon in the first place is that it is used by the etl

src/ot_orchestration/dags/config/platform.yaml
```diff
@@ -0,0 +1,102 @@
+gcs_url: gs://open-targets-pre-data-releases/24.06dev-test-config
+
+pis_version: 24.6.0
```
project-defiant, line 3: Should pis version be in it's own config?

src/ot_orchestration/dags/config/platform.yaml
```diff
@@ -0,0 +1,102 @@
...
+etl_version: 24.1.0
+
+chembl_version: '34'
+efo_version: v3.68.0
+ensembl_version: '112'
+
+is_ppp: false
+
+# At the moment, there is no way to determine the dependencies
```
project-defiant, line 12: By the next best you mean settle the dependencies?  I would say it's superior solution to defining them by comparing hashed inputs | outputs dynamically for instance.

Imagine a situation when we change a significant part of the dag, introduce new dependencies and somehow break the dag. Debugging this without the explicit representation of the topology can be painful. 

This is clean, unless you explicitly link all dependencies. 

The ones that can be co-linked by another node should link to the other node, so we linearize the dag.

src/ot_orchestration/dags/platform.py
```diff
@@ -0,0 +1,192 @@
...
+
+                j = EmptyOperator(
+                    task_id=f"join_{step_name}",
+                    trigger_rule="none_failed_min_one_success",
+                )
+
+                # add the run task to the step registry
+                steps[f"pis_{step_name}"] = j
+                # here we define the task dependencies for both branches
+                c >> Label("invalid previous run") >> u >> r >> d >> j
```
project-defiant: It's mayby just my editor config, but `>>` gives me warnings. I would prefer to use `chain` instead

src/ot_orchestration/dags/config/platform.yaml
```diff
@@ -0,0 +1,102 @@
...
+# of the different steps in the ETL stage. The ETL config file
+# (see https://github.com/opentargets/platform-etl-backend/blob/master/src/main/resources/reference.conf)
+# lacks an explicit definition of dependencies in the step definition.
+# This is the next best option, but it is far from ideal.
+#
+# Any dependency change has to be reflected here _manually_.
+etl_steps:
+  - name: go
```
project-defiant, line 20: Where are the parameters to run the etl steps?

src/ot_orchestration/operators/gce.py
```diff
@@ -572,17 +563,18 @@ def copy_machine_logs(self) -> None:
             page_size=1000,
         )
         for entry in entries:
-            self.log.log(level=logging.INFO, msg=entry.payload.get("message", ""))
+            self.log.info(entry.payload.get("message", "Empty log message"))
```
project-defiant, line 566: What is the reason of this logs?

src/ot_orchestration/operators/gcs.py
```diff
@@ -34,15 +35,14 @@ def __init__(
...
-        # try to get upload url from yaml_file
         self.project_id = project_id
         self.dst = GCSPath(dst)
         self.src = src
 
         self.bucket_name, self.path = self.dst.split()
 
     def execute(self, context) -> None:
-        """Upload the configurations to GCS."""
+        """Execute the Operator."""
```
project-defiant, line 45: This kind of docstring is very trivial. I wonder if we really need them.

src/ot_orchestration/operators/gcs.py
```diff
@@ -51,3 +51,95 @@ def execute(self, context) -> None:
...
+
+class UploadRemoteFileOperator(BaseOperator):
+    """Custom operator that uploads a remote file to GCS.
+
+    This operator will create a GCS bucket if it does not exist and upload the
+    file from a URL to a path inside that bucket.
+
+    Args:
+        project_id: The GCP project ID. Defaults to the platform project.
+        src: The URL to the file to upload.
```
project-defiant:
```suggestion
        src: Source file URL
```

The name of the operator is a bit misleading.

src/ot_orchestration/operators/gcs.py
```diff
@@ -51,3 +51,95 @@ def execute(self, context) -> None:
...
+        self.dst = GCSPath(dst)
+        self.src = src
+
+        self.bucket_name, self.path = self.dst.split()
+
+    def execute(self, context) -> None:
+        """Execute the Operator."""
+        c = Client(project=self.project_id)
+        b = Bucket(client=c, name=self.bucket_name)
+        temp_file = Path("/tmp") / self.src.split("/")[-1]
```
project-defiant, line 90: You could collect all chunks and save without the temporary file in this context.  Or even better fire the `UploadStringOperator` after reading the URI.

src/ot_orchestration/utils/dataproc_platform.py
```diff
@@ -0,0 +1,225 @@
...
+from google.cloud.dataproc_v1 import Cluster, JobReference
+from google.cloud.dataproc_v1.types.jobs import Job, JobPlacement, SparkJob
+
+from ot_orchestration.utils import random_id
+from ot_orchestration.utils.common import GCP_PROJECT_PLATFORM, GCP_REGION
+from ot_orchestration.utils.dataproc import ClusterGenerator
+from ot_orchestration.utils.labels import Labels
+
+
+class DataprocCreateClusterOperator(_DataprocCreateClusterOperator):
```
project-defiant: Convince me that it is a good idea to have the name overwritten.

src/ot_orchestration/utils/dataproc_platform.py
```diff
@@ -0,0 +1,225 @@
...
+            impersonation_chain=self.impersonation_chain,
+            **kwargs,
+        )
+
+    def _create_cluster_config(self) -> dict:
+        # This is the legacy cluster configuration for ETL dataproc clusters.
+        # There are minor changes, but it must be fine-tuned as we start testing
+        # the new pipeline runs.
+        # The only changes so far are:
+        # - The master machine has been downsized to n2-standard-4
```
project-defiant: Ask @prashantuniyal02, but I believe that we were suppose to use `n1` machines.

src/ot_orchestration/utils/dataproc_platform.py
```diff
@@ -0,0 +1,225 @@
...
+            credentials, or chained list of accounts required to get the access_token
+            of the last account in the list, which will be impersonated in the request.
+            If set as a string, the account must grant the originating account
+            the Service Account Token Creator IAM role.
+            If set as a sequence, the identities from the list must grant
+            Service Account Token Creator IAM role to the directly preceding identity, with first
+            account from the list granting this role to the originating account (templated).
+    """
+
+    template_fields: Sequence[str] = (
```
project-defiant: Do you need to overwrite the template_fields?

src/ot_orchestration/utils/labels.py
```diff
@@ -0,0 +1,57 @@
...
+from ot_orchestration.utils.common import GCP_PROJECT_PLATFORM, shared_labels
+
+
+class Labels:
+    """A collection of labels for Google Cloud resources.
+
+    Includes a set of default labels, and ensures that all labels are correctly
+    formatted.
+
+    Refer to the `controlled vocabularies <https://github.com/opentargets/controlled-vocabularies/blob/main/infrastructure.yaml>`__
```
project-defiant, line 14: So the controlled vocab does not contain `genetics` team 🥶

src/ot_orchestration/utils/labels.py
```diff
@@ -0,0 +1,57 @@
...
+        self.label_dict.update({k: self.clean_label(v) for k, v in self.extra.items()})
+
+    def clean_label(self, label: str) -> str:
+        """Clean a label for use in google cloud.
+
+        According to the docs: The value can only contain lowercase letters, numeric
+        characters, underscores and dashes. The value can be at most 63 characters
+        long.
+        """
+        return re.sub(r"[^a-z0-9-_]", "-", label.lower())[0:63]
```
project-defiant, line 44: By looking at this I had to check if slicing string with right index being bigger, then the length of the string will not break it. 🐙  For the sanity I would change it to 
```suggestion
       max_char_count = 63
       rid = max_char_count if len(label) > max_char_count else len(label)
       return re.sub(r"[^a-z0-9-_]", "-", label.lower())[0:rid]
```

project-defiant (approved):
Most of the comments are purely to ensure I understand the overall process. Otherwise approved :)

pyproject.toml
javfg: Nah, it's just the ETL uses HOCON files for config and we have to write one. I'd like to keep that to a minimum and we don't plan to extend its use. If anything, I'd remove it. :)

src/ot_orchestration/dags/config/platform.py
javfg: Are you sure? This python file is where the config for the platform dag comes, effectively. More than parsing it is building it. Besides, I don't think we will have more files like this, so that folder would have only one file.

But if you prefer so, I'm fine with that too.

src/ot_orchestration/dags/config/platform.py
javfg: Yup, or perhaps eventually fetch those from somewhere so they are up to date without the need to copy them here with every change on their respective repos...

Any ideas?

src/ot_orchestration/dags/config/platform.py
javfg: I was not worrying very much about PPP. The mechanism to run its pipeline needs some reevaluation altogether.

In the current process we use at the moment for releases,  there are some different steps: we copy files from the platform run of PIS instead of running it again so they are exactly the same.

Maybe we can have two copies of the dag altogether? As I said, this is just an initial approach but we have to think it through.

project-defiant:
@javfg align the PR name to the semantics

src/ot_orchestration/dags/config/platform.py
project-defiant: Like I mentioned, just for the convention, I will not make a big noise around if you prefer it to stay here :)

src/ot_orchestration/dags/config/platform.py
project-defiant: I wish we do not move it anywhere, as the development process will be difficult then, but on the other hand if we move them to separate repository we could version them with the platflorm releases (for instance). Here we could do that too, but the release process will be affected by the code changes

src/ot_orchestration/dags/config/platform.py
project-defiant: Lets think about it later.

src/ot_orchestration/dags/config/platform.py
javfg: Yup, I find it confusing too. We need to write some design docs detailing all these things. So I managed to differentiate by splitting things:

* Airflow context:
  - dag -> task_group -> task (node?)
  - run (a single run of a dag) -> try (a single run of a task inside of a dag run)

* PIS context:
  - step (e.g.: `baseline_expression`) -> task (download a file)

javfg: I forgot to add: a step of PIS becomes a task in airflow.

src/ot_orchestration/dags/config/platform.py
javfg: Yup, but as I said, I would not tie more config formats into the code, so we keep it into a minimum and get rid of it ASAP. :D

src/ot_orchestration/dags/config/platform.yaml
javfg: I intended this as the DAG config, in fact. So you tell it which version of PIS it should run. I tried to keep the config to a minimum, so in the process of making a new pipeline run, you only see in the config what you _explicitly_ need to be mindful of:

- GCS bucket where to put the output
- Software versions: (PIS version, ETL version)
- Data versions: (EFO, CHEMBL, EMSEMBL versions)
- Whether it is a PPP run or not.

There may be more stuff to configure for the Gentropy side?

src/ot_orchestration/dags/config/platform.yaml
javfg: My idea originally was for the DAG to be able to infer them without explicitly describing them. This is possible, because you know the files each PIS step produces (it it stated in [its config](https://github.com/javfg/platform-input-support/blob/master/config.yaml#L21)), and you _should_ know what files does [each ETL step needs](https://github.com/opentargets/platform-etl-backend/blob/master/src/main/resources/reference.conf#L87).

The problem is the config file for the ETL does not have a rigid structure for defining the inputs, so it is very hard to figure them out right now. Maybe with some changes or clever code, but meanwhile I just define them here.

There is a great upside of not defining the step dependencies but having them inferred from the way PIS and ETL work: any change in the deps will be reflected in the DAG automatically! Saving the hassle and mistakes of introducing new deps but forgetting to detail them in this file.

We should try to achieve that.

src/ot_orchestration/dags/config/platform.yaml
javfg: In the [ETL config file](https://github.com/opentargets/platform-etl-backend/blob/master/src/main/resources/reference.conf). Airflow is just in charge of saying:

```
Hey ETL version xyz, run the step abc for me and drop the output in bucket jkl
```

ETL should know what to do with that. That way we separate concerns.

src/ot_orchestration/operators/gce.py
javfg: Some times google cloud generates empty log messages (maybe if there is a newline somewhere). It should not happen in PIS, but just in case.

src/ot_orchestration/operators/gcs.py
javfg: We don't, I agree completely. But pre-commit complains if I remove it and I didn't find a way to add ignores by method name. :cry:

src/ot_orchestration/operators/gcs.py
javfg: But then if I download a 300GB file, I'd need a 300GB buffer. The chunking is there to lower memory requirements.

Regarding the second part, I think to use an operator inside another operator you have to subclass it. In this particular situation it increases the complexity and makes it way harder to follow (you have to go to the parent to check what's done), and the shared part is just four lines: create a bucket and upload the file. I'd argue it is better to repeat them in this case.

src/ot_orchestration/utils/dataproc_platform.py
javfg: It's not, I was doubtful about it :D

src/ot_orchestration/utils/dataproc_platform.py
javfg: I need to change them to specify the fields I want templates to work on. I don't overwrite them in the parent class though, the parent (`DataprocSubmitJobOperator`) will still have its own template_fields class property.

src/ot_orchestration/utils/labels.py
javfg: AFAIK it's extremely work in progress. I only saw it used in one place before, but I think it's a good idea. You should open a PR for it, I already added some stuff to it myself. :D

src/ot_orchestration/dags/config/platform.py
project-defiant: Could you add a note about it in the docs?

src/ot_orchestration/dags/config/platform.yaml
project-defiant: I believe so, there will possibly be much more I am afraid 🗡️

src/ot_orchestration/dags/config/platform.yaml
project-defiant: With the current approach, even if you forget to add a dependency to the config - it will still be shown in the DAG as a separate node and correct the dependencies list to fit the node just where is should be. 

The only comment why I do not think the dynamic graph dependency is going to fail are:
1. when you have to inject the step in between the existing config, that can not be generated by ingesting the config (like some generic operator)
2. When dag parsing fails to collocate the two dependencies due to discrepancies in between the step input | outputs. You will have to still know where the step should fill in.
3. In gentropy_etl, even if we structurize the parameters and sort them with regards to inputs and outputs, we will still end up with different parameters in dependent steps:
```yaml
steps: 
    - step1:
       output: gs://bucket/datasets/dataset1
    - step2:
       output: gs://bucket/datasets/dataset2
    - step3:
       input: gs://bucket/datasets
```
This will be effectively not possible to figure out the dependency without strictly linking them together.

lets park this for now and see after the `genetics_etl` integration how to fill in.

src/ot_orchestration/utils/labels.py
javfg: I'm not sure we need that check, the [behavior of slices](https://github.com/python/cpython/blame/bd250300191133d276a71b395b6428081bf825b8/Objects/sliceobject.c#L126) has been there for a long time (26 years). It's safe to assume it will not change.

You can even do 
```python
something[3:2]
```
although I'm not sure why would you. :D

src/ot_orchestration/operators/gcs.py
project-defiant: Let's keep in mind that and try to resolve in future.

src/ot_orchestration/operators/gcs.py
project-defiant: You convinced be about skipping DRY.

If you keep 300GB payload, eventually you have to allocate as much fs memory. In that case it would be better to just transfer the chunk directly to the blob in one single loop. This way you do not need the memory for storing the temporary file as well.

src/ot_orchestration/utils/dataproc_platform.py
project-defiant: Let's keep it for now, there should not be any issues.

src/ot_orchestration/dags/config/platform.py
javfg: I'm planning to write a design doc, or at least a design paragraph :D in the README.

src/ot_orchestration/operators/gcs.py
javfg: Indeed, but for that we need to write our own streaming method to upload to google cloud. Unless there is something in Google Cloud API to upload a stream.

src/ot_orchestration/utils/dataproc_platform.py
javfg: I changed it in the suggestions commit. :+1:

PR #26, opened 2024-09-23
refactor: genetic dags

### LINKED ISSUE #27: Unify existing gentropy DAGs
## Objective

As a developer I want to **unify the way we create** DAGs and define the topology for existing genetic data sources processing steps. 

### Context

Genetic DAGs use the `gentropy steps` to perform specific actions releated to genetic data source processing. Datasources ingested by genetic dags consist of:
- finngen GWAS fine mapping results
- finngen ukbppp meta analysis results
- ukbppp eur QTL results
- gnomad variants
- gnomad ld_matrix
- gwas_catalog top associations
- gwas_catalog summary_statistics
- eqtl_catalogue QTL fine mapping results

Most of the sources above generate from 1 to 2 dags (except gwas_catalog, in which the DAGs unify both sources before processing in continuous manner).

All of the genetic_dags (except gwas_catalog releated ones) have specific life cycle coupled with datasource release process that is not handled by Open Targets, so they are not updated during each data release, rather only when the datasource has an update itself (for example bump of finngen GWAS release from r10 to r11).

The entire business logic to full genetic processing is defined in three places:
- [gentropy](https://github.com/opentargets/gentropy/tree/dev/src/gentropy) - contains business of each of the DAGs tasks within the `gentropy step classes` that are exposed by hydra command line interface. 
- [hydra structured configuration](https://github.com/opentargets/orchestration/tree/dev/config) - contains all parameters that define how each of the tasks should be called inside the DAGs
- [dags](https://github.com/opentargets/orchestration/tree/dev/src/ot_orchestration/dags) - contains topologies of each specific datasource processing and overwrites for the hydra configuration

> [!IMPORTANT]
> Each of the gentropy steps have to be run on `spark cluster`, currently the dags allow for invoking the steps inside the dataproc cluster, by copying entire hydra configuration to the dataproc cluster, while overwriting some of the parameters directly inside the DAG.

While this process is easy to follow when developing new steps, understanding entire process requires heavy amount of following nested configuration that can be defined in multiple locations (gentropy config.py, structured config, overwrites in the DAGs). 

The entire process is also error prone to dependency overwrite issue, as the cluster configuration and gentropy version are decoupled from the DAGs and are pre-requisite run by `make build` command inside the gentropy itself. This way the data generated by the DAGs does not contain any relevant information on which version of gentropy was used to generate it.

## Tasks

- [x] The DAGs topology should be inferred directly from the dag configuration, as it is done inside the `gentropy_etl.py` dag 
- [ ] The DAGs should not overwrite the parameters introduced by confuration layer
- [ ] For each DAG , it's step configurations should be extracted from the overall hydra configuration and wrapped around a single config file - this will streamline the readability of the dags.
- [ ] Allow for specifying cluster configurations (startup script, specific gentropy version and dependency installer) via the dag configuration.

## Next steps
- [ ] gwas_catalog dags should be deprecated, instead the `gwas_catalog_pipeline` dag should be promoted as a production ready dag

project-defiant:
## Context

This PR closes https://github.com/opentargets/orchestration/issues/27

The aim of this PR is to **unify existing dags** (except the genetics_etl) to reuse existing approach for `generate_dag` logic implemented for the `genetics_etl` that creates the topology of the dag based on configuration file.  

This process streamlines the dependency management and allows for better understanding of the dependencies between the DAG steps.

Previous implementation had configuration distributed accross multiple files in the [`config` directory](https://github.com/opentargets/orchestration/tree/dev/config/step). This way the configuration was not isolated for each DAG, resulting in heavy lookup into the nested structures of the configs and dags code to understand the overall processes. 

By merging configuration of multiple gentropy steps and extracting this config as a single entity called `dag config` that is stored under the `src/ot_orchestration/dags/config/*.yaml` should increase the readability and verbosity of each process.
Enabiling the nodes and prerequisites in most cases allows to skip on reading the logic of the DAG itself and focus on the process definition maintained in the `dag config`.

## Things implemented:
1. Refactoring of `ukb_ppp_eur_harmonisation` DAG
2. Refactoring of `gwas_curation_update` DAG
3. Refactoring of `gwas_catalog_preprocess` DAG
4. Refactoring of `gnomad_ingestion` DAG
5. Deprecation of `gwas_catalog_harmonisation` DAG -> the content is under development of `gwas_catalog_pipeline` DAG
6. Refactoring of `finngen_ukb_meta_harmonisation` DAG
7. Refactoring of `finngen_ingestion` DAG + addition of extra parameter `sample_size`
8. Refactoring of `eqtl_ingestion` DAG.
9. New bunch of tests for utils
10. Refactoring of `dataproc` releated functions.
11. Fixes to development process 
  - Allow for shell to be inferred from env variable, so after running `make dev` the bashrc file is not populated with junk lines, 
  - Remove sourcing of poetry shell as default from setup script, 
  - Fix duplicate pre-commit call that causes the pre-commit to run twice


project-defiant:
@javfg ready to review!

src/ot_orchestration/dags/eqtl_ingestion.py
```diff
@@ -0,0 +1,72 @@
...
+        cluster_name=config["dataproc"]["cluster_name"],
+        step_name=step_config["id"],
+        python_main_module=config["dataproc"]["python_main_module"],
+        params=step_config["params"],
+    )
+
+    delete_decompressed_job = GCSDeleteObjectsOperator(
+        task_id="delete_decompressed_files",
+        bucket_name=config["eqtl_catalogue_decompressed_susie_path"].split("/")[2],
+        prefix=f"{config['eqtl_catalogue_decompressed_susie_path'].split('/')[-1]}/",
```
javfg: Couldn't we use the `path.split()` method you implemented in your Path class here?

javfg (approved):
It all looks pretty good, except for that little comment.

I'm starting to worry a bit about the organization of the code for this part, though. 

I'll explain:

It seems to me we're extracting all the structure of the DAGs into external functions and utilities, and I wonder how good that is. I know it saves repetition, but I am not against repeating certain things if that means the DAG structure is more explicit. I feel like repetition could even be a good thing.

I'm still not knowledgeable enough about the genetics DAGs so I assume this is the way to go. But to me it looks a bit strange that although we've extracted most of the structure from the python files, we still need one for each and we don't get the benefit of making it flexible enough that we can have a single DAG that just runs config files. Not sure we would even want that (I think I don't). :)

In any case, this was just food for thought. Good job! :tada:

src/ot_orchestration/dags/eqtl_ingestion.py
project-defiant: Done

javfg (approved):

PR #29, opened 2024-09-24
feat: variant-index refactoring

### LINKED ISSUE #3431: Unify Gentropy ETL and Variant Index DAGs
As a **developer** I want **to define all the Gentropy steps required to produce a data release in a single DAG** because **we currently have dependencies between DAGs that require manual intervention**.

## Background

This is the current `genetics_etl` DAG.
image

The actual DAG that calls Gentropy steps and sets their dependencies was originally defined using a YAML file.
```
# for example
- id: "ot_variant_to_gene"
  prerequisites:
    - "ot_gene_index"
```
The DAG reads this YAML and dynamically creates the tree of nodes. This solution worked because all the steps in this process were actual Gentropy steps that were submitted using the Spark Job operator.

The new version of the variant index has introduced complexities that have required to develop an independent DAG, that calls several types of operators:
image

**This breaks the independence between DAGs because:**
1. To call the variant index DAG, you need to produce credible sets in the genetics ETL DAG
2. The variant index is a dependence of the V2G step in the genetics ETL DAG

## Tasks

We need to solve this circularity issue by integrating the variant and the genetics DAGs into a single one

- [ ] Ditch `dags/configs/dag.yaml`
- [ ] Update `dags/genetics_etl.py` so that the Dataproc jobs are instantiated in this module
- [ ] Define a task group that stores the logic from `dags/variant_index.py` and move it to `dags/genetics_etl.py`

project-defiant:
## Context

This PR closes https://github.com/opentargets/issues/issues/3431

image

The main development being done was moving the code from [variant_index dag](https://github.com/opentargets/gentropy/blob/6469bf5dc4311675a90ae60e3ed1a8f0b4da9349/src/airflow/dags/variant_index.py) to the genetics_etl dag to port new tasks (wrapped into `variant annotation` task group) that are required to run before the `variantIndexStep` can be successfully executed.

## Things implemented:

- [x] Operator for annotating variant source with VEP command
- [x] Operator for partitionning the variant sources into chunks required by VEP batch job
- [x] TaskGroup in the configuration
- [x] Addition of biosample index step
- [x] New filters for validation steps  
- [x] Deprecation of old hydra configuration 

## Limitations

The orchestration of `variant_index` and `variant_annotation` currently runs as:

1. convert variant source to vcf (batch job) - 4 tasks | ~ 4min
2. sort and partition vcfs into small chunks (python operator) - partitionned to 850 vcfs | **very heavy in computations on local machine that runs airflow and requires pandas as a dependency** ~ 3min
3. annotate partitionned vcfs (batch job) - 850 tasks ~ 25 min
4. perform variantIndexStep (merge variants from gnomad and variant sources) ~25min

The steps 1 and 2 should be merged and run as a dataproc job (the 1. requires only 4 batch tasks) for silplifying the configuration,  although this is out of scope of this PR. 

Test perfomed on successful run of the pipeline ending at the variant_index step (including) resulted in ~1h10min run


javfg (approved):
I think this is a good start!

I haven't gone in detail into it. As we discussed yesterday, we should try to figure out a way to separate concerns in a similar way as it is being done in the platform, with a [configuration file](https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/unified_pipeline.yaml) for the pipeline itself and [other files](https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/pis.yaml) for the settings for the internals of running an application inside a step.

We could do this by either overriding settings in a hierarchical way (tool setting < pipeline settings), by templating, or by trying to isolate settings in one of the two places.

I realize this is harder in Gentropy due to it being cloud-agnostic. That is a target for the platform part too. Even if it's impossible to do completely, we should strive for it. I think this separation will enable us to separate concerns in the orchestrator code too.

PR #31, opened 2024-09-30
fix: remove lut and unpablished from gwas_curation_update

addramir:
Fix to unblock the curation table production dag.


d0choa:
I'll try to handle this in #33

PR #32, opened 2024-10-01
feat: allow for force reinstall dependencies on cluster when set up

project-defiant:
# implementations

Added step to force reinstall the dependencies in the dataproc cluster to ease the development on already set up cluster.

This feature just reenables to reinstall the `gentropy` package with force flag enabled in the `install_dependencies.sh` script. - see https://github.com/opentargets/gentropy/pull/804. This feature will have to be manually added in case it's needed in the dag.


d0choa (approved):
Pending testing but it looks good to me.

PR #36, opened 2024-10-04
refactor: clarify paths

d0choa:
Updated paths from study_locus to study_locus_window_based_clumped following Yakov's advice


d0choa (commented):
I have updated the paths accordingly in GCP

PR #39, opened 2024-10-11
feat(ukb_ppp_eur) dags refactoring

project-defiant:
# Context

This PR closes [#3493](https://github.com/opentargets/issues/issues/3493)
We want to be able to run the harmonisation and perform susie finemapping batch job within the orchestration for ukb_ppp_eur data. 

This PR summarizes the developments over the processing of the
- ukb_ppp_eur_harmonisation (harmonisation + locus breaker) that results in the study_locus dataset partitioned by studyLocusId (run via dataproc)
- ukb_ppp_eur_finemapping (finemapping manifest generation + susie finemapping) that results in the credible_set dataset partitioned by studyLocusId. (run via google batch)

To run the harmonisation, some steps needs to be pre-executed before. I have described these steps in the docs along with the data structure.

## Additional tasks
- [x] Consolidated structure for batch operators (VEP and fine-mapping) in the package.
- [x] New function to extract `partition` from path.
- [x] Additional parameters `AllocationPolicy` added to the `create_batch_task` function.
- [x] Cleaned old data in `ukb_ppp_eur_data` and uploaded new datasets from last successful run of the dags.
- [x] Updated README documentation for the `ukb_ppp_eur_data` bucket.
- [x] Added a brief description of the fine-mapping step
- [x] Fixed issue with `dataproc` jobs that used the `local[*]` by default instead of `yarn`. 
- [x] Readded the idle timeout for cluster deletion.
- [x] Added `graphviz` as a dev dependency to generate dag svgs.
- [x] Added instruction on how to generate dag svg.   

> [!NOTE]
> The `yarn` configuration option will now be added automatically to the dataproc steps

PR #40, opened 2024-10-14
fix(yarn): add dataproc parameter to hydra comamnd parsing function in gentropy step

project-defiant:
# Context

Dataproc job are using `local[*]` in the after the https://github.com/opentargets/orchestration/pull/39. This PR updates the `convert_params_to_hydra_positional_args` invocation to include `dataproc=True` so the `yarn` flag is ammended it it is not present in the dataproc config.


d0choa (approved):

d0choa:
@ireneisdoomed ran into the issue today when running eQTL catalogue DAG. Hopefully, this will sort out the issue permanently

PR #45, opened 2024-10-16
feat: working genetics_etl including L2G

d0choa:
This configuration produced `24.10_freeze2` against `szsz-nonmandatory-params-to-coloc-step`.

I commented `ShortCircuitOperator` to make this run because it wasn't working for me, but I don't know if I used any wrong assumptions.


project-defiant:
@d0choa something with pre-commit is issuing

PR #47, opened 2024-10-17
chore(finngen_ingestion): update configuration

project-defiant:
# Context

We want to bring back the `FINNGEN_R11_` prefix to the `studyId` fields and remove duplicates from the credible sets that were read by providing too wide pattern (capturing credible sets and 99 credible sets)


d0choa (approved):
Fingers crossed

PR #52, opened 2024-10-21
feat(add_efo_biosample): add input to orchestration

Tobi1kenobi:
Just adding a path to the efo.json to allow for cell lines in the biosample index


DSuveges (approved):
New parameter added to biosample index generation step.

PR #60, opened 2024-10-28
feat(freeze_5): november data release freeze5

project-defiant:
## Context

Changes added to the configuration
- clean credible sets from gwas catalog sumstats susie and ukb_ppp_eur
- cahnged release path to freeze5
- use the studyIndex from top hits instead of pics


DSuveges (approved):
All makes sense to accommodate to the new release. In the next iteration, we'll change the study index paths so all three sources from GWAS Catalog will be considered.

PR #65, opened 2024-11-01
chore: name l2g run with release id and move download_from_hub param

ireneisdoomed:
Each L2G run ID in W&B will correspond with the release ID. Also `download_from_hub` is a param of the predicting step.


project-defiant (approved):

project-defiant:
Make so much sense. Thanks for it!

PR #66, opened 2024-11-01
feat: updating invalid qc reasons for validation

DSuveges:
- Adding new study index sources
- Updated list of QC flags for study and credible set exclusion


DSuveges:
PR depends on Gentropy [PR#887](https://github.com/opentargets/gentropy/pull/887)

PR #67, opened 2024-11-04
feat: refactoring vep annotation step

project-defiant:
# Context

This PR is a part of effort to reduce the complexity of genetics_etl dag. Due to the fact that we want to submerge the genetics_etl to the platform etl dag, we need to simplify the VariantAnnotation task group.

## Implementations

This PR changes the logic of `variant_annotation` node in the genetics etl dag.

- [x] Change the  `variant_to_vcf` dag node implementation:
  Node was previously running as a google batch job with  4  `variant_to_vcf` gentropy steps (one per each datasource
      * uniprot
      * eva
      * credible sets
      * pharmgkb)  
   that was followed by `local` (run with python operator - `ConvertVariantsToVcfOperator`) partitioning of the vcf based data sources.
   With refactoring of `variant_to_vcf` step implemented in  https://github.com/opentargets/gentropy/pull/891 the step can now take **n datasources** that are running in sequence and produce a single partitioned dataset. This allows for the:
* removal of google batch task for variant to vcf convertion
* removal of the local partitioning of vcf data sources
* drop pandas as dependency (no longer needed, as `ConvertVariantsToVcfOperator` is no longer there
* removal of the entire TaskGroup and adding a single dataproc node for variant annotation instead
- [x] Refactoring of the VepPathManager and addition of the tests
- [x] Preparation for freeze_6
- [x] Handling issue with spark not able to write csv with `#` in the column name without quoting the column name - issue comes when ` variant_to_vcf` wants to save the `#CHROM` column as `"#CHROM"` creating the following issue in VEP:
```bash
WARNING: line 1 skipped ("#CHROM" POS ID REF ALT QUAL FILTER INFO): Invalid start 'POS' or end '2' coordinate
```

> [!NOTE]
> This PR requires the https://github.com/opentargets/gentropy/pull/896 to be merged, as the `#CHROM` issue was resolved by dynamically change the `CHROM` to `#CHROM`, so the `variant_to_vcf` step should drop the hash.

### DIFF
--- a/src/ot_orchestration/operators/batch/vep.py
+++ b/src/ot_orchestration/operators/batch/vep.py
@@ -7,8 +7,10 @@ from functools import cached_property
 from pathlib import Path
 from typing import Sequence, Set
 
+from airflow.exceptions import AirflowException
 from airflow.providers.google.cloud.hooks.cloud_batch import CloudBatchHook
 from airflow.providers.google.cloud.operators.cloud_base import GoogleCloudBaseOperator
+from google.cloud.batch import JobStatus
 from google.cloud.batch_v1 import Job
 from google.cloud.storage import Client
 
@@ -92,6 +94,10 @@ class VepAnnotateOperator(GoogleCloudBaseOperator):
     This operator performs the VEP annotation of vcf files provided in the `vcf_input_path`.
     The annotation command is custom to OTG needs. The number of batch tasks is inferred by listing the number of
     vcf files introduced in the `vcf_input_path`. Each input file will result in a single vep task.
+
+    The operator tries to list the tasks after they are completed, in case of any failures in the task, the
+    operator throws AirflowException and stops the execution of the DAG. The defails of the failed job can be
+    retrieved from the logs.
     """
 
     def __init__(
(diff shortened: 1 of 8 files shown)


src/ot_orchestration/operators/batch/vep.py
```diff
@@ -205,134 +248,3 @@ def _vep_command(self) -> list[str]:
                 --plugin AlphaMissense,file={self.pm.cache_dir}/AlphaMissense_hg38.tsv.gz,transcript_match=1 \
                 --plugin CADD,snv={self.pm.cache_dir}/CADD_GRCh38_whole_genome_SNVs.tsv.gz",
         ]
-
-
-class ConvertVariantsToVcfOperator(BaseOperator):
```
DSuveges, line 210: As I understand,this is no longer needed, because this functionality is done by a separate step, right?

project-defiant: This is no longer needed, as it is a part of the `variant_to_vcf` step -> https://github.com/opentargets/gentropy/blob/4d8e7c42e8e94473495d12d50efb7d7942afbe73/src/gentropy/variant_index.py#L114

DSuveges (approved):
As far as I understand the bits, all the changes are well thought out. And looks good. However I haven't tested them.

PR #69, opened 2024-11-06
feat: freeze 6 improvements

project-defiant:
# Context:

- [x] genetics etl config changes to accomodate freeze6  rerun for l2g steps
- [x] Improvement over the hadling of variant annotation step in genetics etl

The change to `find_node_in_config` function allows for skipping the step, if config is not present without issueing error when generics ETL dag is parsed, but config does not contain the `VariantAnnotation` node.

### DIFF
--- a/src/ot_orchestration/dags/config/genetics_etl.yaml
+++ b/src/ot_orchestration/dags/config/genetics_etl.yaml
@@ -208,7 +208,7 @@ nodes:
       step.feature_matrix_path: gs://ot_orchestration/releases/24.10_freeze6/locus_to_gene_feature_matrix
       step.credible_set_path: gs://ot_orchestration/releases/24.10_freeze6/credible_set
       step.download_from_hub: true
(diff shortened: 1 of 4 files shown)


src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -208,6 +209,7 @@ nodes:
       step.credible_set_path: gs://ot_orchestration/releases/24.10_freeze6/credible_set
       step.download_from_hub: true
       step.hf_hub_repo_id: opentargets/locus_to_gene
```
ireneisdoomed, line 211: This should be changed if you want this config to be an accurate representation of the dag, right?
```suggestion
      step.hf_hub_repo_id: opentargets/locus_to_gene_egl
```

ireneisdoomed (approved):
Nice, thanks! So this implies that to skip a step we only need to comment the YAML, and not the DAG itself. Is that correct?

project-defiant:
Yes

PR #71, opened 2024-11-07
feat: configuration for etl run with gentropy

project-defiant:
# Context

We want to be able to run the genetics etl dag with two modes:
 - standalone
 - as a part of unified pipeline

This PR attempts to create a configuration for the unified pipeline run with genetices etl.

## Things implemented
- [x] Addition of `biosamples` step to the PIS config (required for `biosample_index` step in genetics etl)
- [x] Set up input and output paths to match the platform release path
- [x] Update of etl_search dependencies 
- [x] Set up input dependencies between (platform, PIS) -> genetics ETL 
  -  `etl_target` -> `genetics_gene_index`
  -  `pis_biosamples` -> `genetics_biosample_index`
  -  `etl_diseases` -> `genetics_study_index`
  -  `pis_evidence` -> `genetics_variant_conversion`
  - `pis_pharmacogenomics` -> `genetics_variant_conversion`
  - `etl_interactions` -> genetics_l2g_train`
- [x] Set up output dependencies between (gentropy ETL) -> platform ETL 
   - `etl_evidence` -> requires `gentropy_evidence` to be stored under the `pis_evidence` output directory
   - `etl_search` -> requires `gentropy_variant_index`

### DIFF
--- a/src/ot_orchestration/dags/config/genetics_etl_platform.yaml
+++ b/src/ot_orchestration/dags/config/genetics_etl_platform.yaml
@@ -44,11 +44,12 @@
 # gs://open-targets-ppp-releases/${release}/input/evidence-files/l2g.json.gz
 
 release: 24.12
+commit: 93de44855840d30bf4bfd5a7dce444adf739ace0
 dataproc:
-  python_main_module: gs://genetics_etl_python_playground/initialisation/gentropy/dev/cli.py
+  python_main_module: gs://open-targets-pre-data-releases/24.09/gentropy/${commit}/gentropy/dev/cli.py
   cluster_metadata:
-    PACKAGE: gs://genetics_etl_python_playground/initialisation/gentropy/dev/gentropy-0.0.0-py3-none-any.whl
-  cluster_init_script: gs://genetics_etl_python_playground/initialisation/gentropy/dev/install_dependencies_on_cluster.sh
+    PACKAGE: gs://open-targets-pre-data-releases/24.09/gentropy/${commit}/gentropy-0.0.0-py3-none-any.whl
+  cluster_init_script: gs://open-targets-pre-data-releases/24.09/gentropy/${commit}/install_dependencies_on_cluster.sh
   cluster_name: otg-etl
   autoscaling_policy: otg-efm
   allow_efm: true
--- a/src/ot_orchestration/dags/config/genetics_etl_platform.yaml
+++ b/src/ot_orchestration/dags/config/genetics_etl_platform.yaml
@@ -1,4 +1,4 @@
-# Inputs:
+# PIS & ETL Inputs:
 # gs://open-targets-pre-data-releases/${release}/output/etl/parquet/targets (etl_target)
 # gs://open-targets-pre-data-releases/${release}/input/biosamples/cl.json (pis_biosamples)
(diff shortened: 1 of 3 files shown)


src/ot_orchestration/dags/config/genetics_etl_platform.yaml
```diff
@@ -0,0 +1,283 @@
...
+# gs://open-targets-pre-data-releases/${release}/output/genetics/parquet/l2g_feature_matrix
+# gs://open-targets-pre-data-releases/${release}/output/genetics/models/locus_to_gene_model/classifier.skops
+# gs://open-targets-pre-data-releases/${release}/output/genetics/parquet/l2g_predictions
+# gs://open-targets-ppp-releases/${release}/input/evidence-files/l2g.json.gz
+
+release: 24.12
+dataproc:
+  python_main_module: gs://genetics_etl_python_playground/initialisation/gentropy/dev/cli.py
+  cluster_metadata:
+    PACKAGE: gs://genetics_etl_python_playground/initialisation/gentropy/dev/gentropy-0.0.0-py3-none-any.whl
```
DSuveges: Maybe not immediately, but these paths should also be phased out. The release folder has place for the code as well: eg `gs://open-targets-pre-data-releases/24.09/jars`

DSuveges: At some point it would be beneficial to tight the data and the software release.

src/ot_orchestration/dags/config/genetics_etl_platform.yaml
```diff
@@ -0,0 +1,283 @@
...
+  num_workers: 10
+
+nodes:
+  - id: genetics_gene_index
+    kind: Task
+    prerequisites:
+      - etl_target # ETL step required
+    params:
+      step: gene_index
+      step.target_path: gs://open-targets-pre-data-releases/${release}/output/etl/parquet/targets
```
DSuveges, line 65: This is not looking good to me. At the end, in the same bucket there will be two datasets named as `targets` and`target_index`. I find it quite confusing. Do we really need both datasets? What is the difference? I know, this question is beyond the scope of this PR.

src/ot_orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -100,3 +100,4 @@ etl_steps:
   - name: search
     depends_on:
       - etl_association
+      - genetics_variant_index
```
DSuveges, line 106: Might out of scope, but the `evidence` step depends on `genetics_l2g_evidence` step.

src/ot_orchestration/dags/config/genetics_etl_platform.yaml
project-defiant: I agree, let's drop the step when we move the `gene_index` processing from gentropy to the ETL directly

src/ot_orchestration/dags/config/unified_pipeline.yaml
project-defiant: you are right! thanks

Edit:
The `associations` require the `evidence` which is a direct successor of `psi_evidence` and `genetics_l2g_evidence`. I have updated the config with this in mind, not to pollute the topology

DSuveges (approved):
Wow, this is a very comprehensive effort.

PR #76, opened 2024-11-12
feat(docs): add command to update the documentation in buckets

project-defiant:
# Context

We want to have the documentation in sync between the google cloud buckets and the code, this PR attempts to create a utility command that will synchronise the documentation to correct buckets.

## Things implemented

- command for updating the GWAS Catalog documentation in the bucket.
- command to update all data source documentations in their corresponding buckets.


DSuveges (approved):
All good.

PR #78, opened 2024-11-18
fix: gnomad variant index and ld preprocess

DSuveges:


src/ot_orchestration/dags/config/gnomad_ingestion.yaml
```diff
@@ -1,32 +1,25 @@
...
-        - afr # African-American
-        - amr # American Admixed/Latino
-        - asj # Ashkenazi Jewish
-        - eas # East Asian
-        - est # Estonian
-        - fin # Finnish
-        - nfe # Non-Finnish European
-        - nwe # Northwestern European
-        - seu # Southeastern European
+      step.ld_index_out: gs://genetics_etl_python_playground/static_assets/ld_index
```
project-defiant, line 16: Can we store it in a data bucket?

project-defiant (approved):

PR #79, opened 2024-11-19
fix: adding NOT_QUALIFYING_LD_BLOCK to exclusion

addramir:
Related to bug when credible set size is zero for 10445 CSs.
The problem is that r2 for the lead variant is 0 for 100% cases.
There are two possible fixes:

1. Ensure that r2 for lead is always 1 in the ld annotation step.
2. Add the flag "[LD block contains no variants at the required R^2 threshold]" as an exclusion criterion.

This PR fixes 2).


project-defiant:
Closing as the `NOT_QUALIFYING_LD_BLOCK` is not in the validation rules anymore.

PR #84, opened 2024-11-25
Unified Pipeline

javfg:
This PR adds the first (almost) working version of the unified pipeline DAG.

There is a lot of stuff still pending, but we are able to run the whole PIS->Ontoform->ETL->Gentropy->ETL process, except for a few steps that require manual intervention:

* `etl_disease` — pending implementation of hpo and hpo_phenotypes in Ontoform
* `gentropy_variant_annotation` — service account problems in google cloud
* both `gentropy_coloc` steps — pending fixes in Gentropy

I'll make a short summary of the changes in the PR to make it easier to review:

`src/ot_orchestration/dags/config/etl.conf` — ETL config file (with some path changes), no new to review.
`src/ot_orchestration/dags/config/gentropy.yaml` — Gentropy config file. This is the part of the genetics_etl.yaml that contains the configuration for Gentropy itself (more on this below).
`src/ot_orchestration/dags/config/pis.yaml` — Pis config file, no need to review.
`src/ot_orchestration/dags/config/unified_pipeline.py` — This is the Config class for the Unified Pipeline DAG. It builds a running configuration based on all the config files for each part/
`src/ot_orchestration/dags/config/unified_pipeline.yaml` — This is the configuration file for the Unified Pipeline. The aim is that we only need to edit this file create a new release.
`src/ot_orchestration/dags/unified_pipeline.py` — This is the main file for the pipeline DAG, **the most important part to review**.
`src/ot_orchestration/utils/dataproc.py` — Added a `project_id` parameter. This set of methods must be converted into Operators.
`src/ot_orchestration/utils/utils.py` — Added some naming functions and a HOCON file parser for the ETL config.


javfg:
I'd say we're ready to merge this, before it grows even bigger.

We can keep adding on top later as we do fixes/improvements.

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -127,41 +126,43 @@ nodes:
...
     params:
       step: colocalisation
-      step.credible_set_path: gs://ot_orchestration/releases/24.10_freeze6/credible_set
-      step.coloc_path: gs://ot_orchestration/releases/24.10_freeze6/colocalisation/
-      step.colocalisation_method: Coloc
-      +step.colocalisation_method_params: "{priorc1: 1e-4, priorc2: 1e-4, priorc12: 1e-5}"
+      step.credible_set_path: gs://ot_orchestration/releases/24.11_freeze10/credible_set
+      step.coloc_path: gs://ot_orchestration/releases/24.11_freeze10/colocalisation
+      step.colocalisation_method: ECaviar
+      +step.session.extended_spark_conf: "{spark.sql.shuffle.partitions: '3200', spark.executor.memory: '16g', spark.sql.files.maxPartitionBytes: '25000000', spark.executor.cores: '2'}"
```
project-defiant, line 165: Old configuration

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+      step.coloc_path: '{gcs_url}/output/genetics/parquet/colocalisation'
+      step.colocalisation_method: ECaviar
+      +step.session.extended_spark_conf: "{spark.sql.shuffle.partitions: '4000', spark.executor.memory: '25g', spark.sql.files.maxPartitionBytes: '25000000', spark.executor.cores: '4'}"
+      step.session.write_mode: ignore
+      step.session.output_partitions: 25
+
+  variant_annotation:
+    google-batch:
+      <<: *batch_settings
+      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:{vep_version}
```
project-defiant, line 143: Add the `image` and `entrypoint` to the `batch_settings`

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+    params:
+      step: locus_to_gene_feature_matrix
+      step.credible_set_path: '{gcs_url}/output/genetics/parquet/credible_set'
+      step.variant_index_path: '{gcs_url}/output/genetics/parquet/variant_index'
+      step.colocalisation_path: '{gcs_url}/output/genetics/parquet/colocalisation'
+      step.study_index_path: '{gcs_url}/output/genetics/parquet/study_index'
+      step.gene_index_path: '{gcs_url}/output/genetics/parquet/gene_index'
+      step.feature_matrix_path: '{gcs_url}/output/genetics/parquet/l2g_feature_matrix'
+      +step.session.extended_spark_conf: "{spark.sql.autoBroadcastJoinThreshold:'-1'}"
+      step.session.write_mode: overwrite
```
project-defiant, line 169: drop the `overwrite`

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+      step: locus_to_gene
+      step.run_mode: predict
+      step.predictions_path: '{gcs_url}/output/genetics/parquet/l2g_predictions'
+      step.feature_matrix_path: '{gcs_url}/output/genetics/parquet/l2g_feature_matrix'
+      step.credible_set_path: '{gcs_url}/output/genetics/parquet/credible_set'
+      step.download_from_hub: false
+      step.l2g_threshold: 0.05
+      step.model_path: '{gcs_url}/input/benchmarks_l2g_fm0_v5.1_best_cv_locus_to_gene_model_classifier.skops'
+      step.hf_hub_repo_id: opentargets/locus_to_gene
+      step.session.write_mode: overwrite
```
project-defiant, line 183: Drop the `overwrite`

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+  l2g_evidence:
+    params:
+      step: locus_to_gene_evidence
+      step.evidence_output_path: '{gcs_url}/input/evidence-files/gentropy'
+      step.locus_to_gene_predictions_path: '{gcs_url}/output/genetics/parquet/l2g_predictions'
+      step.credible_set_path: '{gcs_url}/output/genetics/parquet/credible_set'
+      step.study_index_path: '{gcs_url}/output/genetics/parquet/study_index'
+      step.locus_to_gene_threshold: 0.05
+      step.session.write_mode: overwrite
```
project-defiant, line 195: Drop the overwrite

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+        - '{gcs_url}/input/evidence-files/uniprot.json.gz'
+        - '{gcs_url}/input/evidence-files/eva.json.gz'
+        - '{gcs_url}/input/pharmacogenomics-inputs/pharmacogenomics.json.gz'
+        - '{gcs_url}/output/genetics/parquet/credible_set'
+      step.source_formats:
+        - json
+        - json
+        - json
+        - parquet
+      step.output_path: '{gcs_url}/output/genetics/parquet/partitioned_variants'
```
project-defiant, line 115: This is not parquet format, it's tsv (has to be for vep to run the annotation)

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+      step.session.output_partitions: 25
+
+  variant_annotation:
+    google-batch:
+      <<: *batch_settings
+      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:{vep_version}
+      entrypoint: /bin/sh
+    params:
+      vep_cache_path: gs://genetics_etl_python_playground/vep/cache
+      vcf_input_path: '{gcs_url}/output/genetics/parquet/partitioned_variants'
```
project-defiant, line 147: As above, this are tsv files (not parquet)

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -0,0 +1,195 @@
...
+      step.model_path: '{gcs_url}/input/benchmarks_l2g_fm0_v5.1_best_cv_locus_to_gene_model_classifier.skops'
+      step.hf_hub_repo_id: opentargets/locus_to_gene
+      step.session.write_mode: overwrite
+      # step.session.write_mode: ignore
+      step.session.output_partitions: 1
+
+  l2g_evidence:
+    params:
+      step: locus_to_gene_evidence
+      step.evidence_output_path: '{gcs_url}/input/evidence-files/gentropy'
```
project-defiant, line 190: NIT, semantically it is the L2G not gentropy

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -25,54 +25,73 @@ class PlatformConfig:
...
         self.pis_config_local_path = Path(__file__).parent / "pis.yaml"
         self.etl_config_local_path = Path(__file__).parent / "etl.conf"
+        self.gentropy_config_local_path = Path(__file__).parent / "gentropy.yaml"
 
         # These are hardcoded config values that are not meant to change often.
         # It is safe to extract them from here into the config file in case they
         # become more dynamic. Just add them to self in the initialization so they
         # are available.
 
         # fmt: off
```
project-defiant, line 38: We should increase the line-length

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -25,54 +25,73 @@ class PlatformConfig:
...
         self.is_ppp = settings["is_ppp"]
+        self.steps = settings["steps"]
+        self.ppp_steps = [s for s, d in self.steps.items() if d and d.get('ppp_only', False)]
 
         # PIS-specific settings.
-        self.pis_config = self.init_pis_config()
         pis_version = settings["pis_version"]
+        self.pis_config = self.init_pis_config()
         # The base image for PIS, the version tag will be appended from the config file.
         pis_image_base = "europe-west1-docker.pkg.dev/open-targets-eu-dev/platform-input-support-test/platform-input-support-test"
```
project-defiant, line 60: The base image is not dynamic, should be included in configuration IMO

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -25,54 +25,73 @@ class PlatformConfig:
...
         self.pis_pool = 16  # number of parallel workers inside of each PIS step
+        self.pis_disk_size = 150 # The disk size for PIS vms, in GB.
+        # Note: although not all steps need this much space, it is easier to have a
+        # single value for all steps, and the machines are so short-lived that it
+        # doesn't matter much with respect to cost.
+
+        # ONTOFORM-specific settings.
+        ontoform_version = settings["ontoform_version"]
+        self.ontoform_step_list = [s for s in settings["steps"].keys() if s.startswith("ontoform_")]
+        self.ontoform_machine_type = 'n1-standard-32'
```
project-defiant, line 72: Again hard coded infra block. Let's keep it as is for now, collect these things and move it outside to configuration

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -25,54 +25,73 @@ class PlatformConfig:
...
+        self.ontoform_image = f"{ontoform_image_base}:{ontoform_version}"
 
         # ETL-specific settings.
+        etl_version = settings["etl_version"]
         self.etl_config = self.init_etl_config()
         self.etl_config_gcs_uri = f"{self.gcs_url}/output/etl-config.conf"
-        etl_version = settings["etl_version"]
         # The base url for the ETL jar, the version will be replaced in from the config file.
         etl_jar_base = "https://github.com/opentargets/platform-etl-backend/releases/download/v{version}/etl-backend-{version}.jar"
         self.etl_jar_origin_url = f"{etl_jar_base.format(version=etl_version)}"
```
project-defiant, line 83: f"..." not needed here

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -105,26 +124,62 @@ def init_pis_config(self) -> dict[str, Any]:
...
+                "vep_version": self.vep_version,
+            },
+        )
+
+    def gentropy_step(self, step_name: str) -> dict[str, Any]:
+        """Return the config for the gentropy step."""
+        real_step_name = step_name.replace("gentropy_", "")
+        step = self.gentropy_config["steps"].get(real_step_name)
+        if not step:
+            raise ValueError(f"Step {real_step_name} not in gentropy config.")
```
project-defiant, line 184: It can be misleading on what config is referred here, can you paste the path to the exact config used?

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -105,26 +124,62 @@ def init_pis_config(self) -> dict[str, Any]:
...
 
+    def pis_env_vars(self, step_name: str) -> dict[str, str]:
+        """Return the environment variables for a PIS step."""
+        return {
+            "PIS_STEP": step_name.replace("pis_", ""),
+            "PIS_CONFIG_FILE": "/config.yaml",
+            "PIS_POOL": str(self.pis_pool),
+        }
+
+    def ontoform_args(self, step_name: str) -> list[str]:
```
project-defiant, line 135: static method (could be moved to utils)

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -105,26 +124,62 @@ def init_pis_config(self) -> dict[str, Any]:
...
+        return read_yaml_config(
+            self.gentropy_config_local_path,
+            sentinels={
+                "gcs_url": self.gcs_url,
+                "gentropy_version": self.gentropy_version,
+                "vep_version": self.vep_version,
+            },
+        )
+
+    def gentropy_step(self, step_name: str) -> dict[str, Any]:
```
project-defiant, line 179: Overall this class is a bit messy, IMO a bit too many things happen during initialization

src/ot_orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -1,106 +1,221 @@
-gcs_url: gs://open-targets-pre-data-releases/24.06dev-test-config
+---
+# The bucket where the pipeline will work in
+gcs_url: gs://open-targets-pre-data-releases/24.12-uo_test-3
```
project-defiant, line 3: Note (to myself and you) When going to production, it would be nice to store this as a airflow config.

src/ot_orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -1,106 +1,221 @@
...
+      - gentropy_variant_index
+  gentropy_colocalisation_ecaviar:
+    depends_on:
+      - gentropy_variant_index
+
+  gentropy_l2g_feature_matrix:
+    depends_on:
+      - gentropy_colocalisation_coloc
+      - gentropy_colocalisation_ecaviar
+      - gentropy_variant_index
```
project-defiant, line 212: I guess, this is more streamlined (originally this should also require the `gentropy_validation` steps, but given colocalisation or variant_index we require them indirectly

src/ot_orchestration/dags/unified_pipeline.py
```diff
@@ -161,36 +217,107 @@ def etl_cluster_prepare() -> None:
...
+    #
+    # Note: labels are generated but not used yet, pending refactor of cluster
+    #       management functions into operators.
+    # ==============================================================================================
+    if len(config.gentropy_step_list):
+        gentropy_cluster_name = create_cluster_name("gentropy")
+        clusterless_steps = []
+
+        c = create_cluster(
+            cluster_name=gentropy_cluster_name,
```
project-defiant, line 268: Can you just pass the `gentropy_dataproc_cluster_settings`, why storing some of the options here and the rest in the config?

src/ot_orchestration/utils/utils.py
```diff
@@ -37,47 +37,87 @@ def clean_name(name: str) -> str:
...
     """
     config_path = config_path if isinstance(config_path, Path) else Path(config_path)
     assert config_path.exists(), f"HOCON config path {config_path} does not exists"
-    with open(config_path) as config_file:
-        return pyhocon.ConfigFactory.parse_string(config_file.read())
+
+    raw_config = config_path.read_text()
+    if sentinels:
+        for sentinel, replacement in sentinels.items():
+            raw_config = raw_config.replace(f"{{{{{sentinel}}}}}", replacement)
```
project-defiant, line 115: this line looks very nasty

src/ot_orchestration/utils/utils.py
```diff
@@ -37,47 +37,87 @@ def clean_name(name: str) -> str:
...
     """
     config_path = config_path if isinstance(config_path, Path) else Path(config_path)
     assert config_path.exists(), f"YAML config path {config_path} does not exists"
-    with open(config_path) as config_file:
-        return yaml.safe_load(config_file)
+
+    raw_config = config_path.read_text()
+    if sentinels:
+        for sentinel, replacement in sentinels.items():
+            raw_config = raw_config.replace(f"{{{sentinel}}}", replacement)
```
project-defiant, line 79: why not use the str.format directly? it will be less ugly

src/ot_orchestration/dags/config/etl.conf
```diff
@@ -1,10 +1,1574 @@
-common.input = "gs://open-targets-pre-data-releases/24.06dev-test/input"
-common.output-base_path = "gs://open-targets-pre-data-releases/24.06dev-test/output"
+// --- Templated parameters --- //
 
-chembl_version = "34"
-ensembl_version = "112"
+common.input = "{{gcs_url}}/input"
```
project-defiant, line 3: I am no expert to review this conf. Please get feedback from somebody from BE on this

project-defiant (commented):
Massive work @javfg, all makes sense. I have some minor comments, please take a look. Thank you for this!

src/ot_orchestration/dags/config/gentropy.yaml
javfg: The idea here is to have a common set of settings for google batch tasks and append there specific settings, like the image and entrypoint, so you don't have to repeat the whole thing if you ever add another batch task.

src/ot_orchestration/dags/config/gentropy.yaml
javfg: Are you sure about this? It was there originally in your code: https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/genetics_etl_platform.yaml#L234

src/ot_orchestration/dags/config/gentropy.yaml
javfg: Same here: https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/genetics_etl_platform.yaml#L271-L271

src/ot_orchestration/dags/config/gentropy.yaml
javfg: This is copied from the original code you wrote too: https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/genetics_etl_platform.yaml#L163

src/ot_orchestration/dags/config/gentropy.yaml
javfg: Same: https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/genetics_etl_platform.yaml#L185

src/ot_orchestration/dags/config/gentropy.yaml
javfg: I'm not sure I understand this comment. If it means L2G is not part of gentropy, let's assume here gentropy encompasses the whole genetic part of the pipeline.

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: I agree wholeheartedly :laughing: 

120 line length nao!

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: It does not make sense to have a setting for the base image of PIS. This will never change (except when we actually move the new PIS on top of the old one. It is not something we will have to change for newer releases.

The logic here is: this class is also part of the configuration, but we only expose into actual yaml files what should be changed.

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: Another thing that will not change in the foreseeable future. If this is something we would have to ever tinker with, we can take it out.

javfg: See https://github.com/opentargets/orchestration/blob/dev/src/ot_orchestration/dags/config/unified_pipeline.py#L32-L35

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: I don't understand this one. Do you mean in the `etl_version`?

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: Don't you think it belongs better here with the rest of ontoform-config related utilities?

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: Added.

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: Anything specific you thing does not belong into configuration initialization?

src/ot_orchestration/dags/config/unified_pipeline.yaml
javfg: Keep in mind if you do that, you can only access that value during a run. I thought the same until I tried, not sure if it's the best idea tho.

src/ot_orchestration/dags/config/unified_pipeline.yaml
javfg: Yup, transitive dependencies are not explicit: https://github.com/opentargets/orchestration/blob/unified-orchestrator/src/ot_orchestration/dags/config/unified_pipeline.yaml#L27-L29

src/ot_orchestration/dags/unified_pipeline.py
javfg: the cluster name is dynamic (per run), and based on the task group name. I could move it to the config, but are you sure it makes sense?

src/ot_orchestration/dags/config/etl.conf
javfg: No need to review this, it's a copy of the original. The orchestration is happening here now, so the etl config belongs here.

src/ot_orchestration/utils/utils.py
javfg: It's not that easy, because there are other strings that look like sentinels in the file, but are not meant to be replaced. [The way `str.format()` works](https://docs.python.org/3/library/stdtypes.html#strw.format) is, it looks for all the strings in braces in the sentinels dict.

In our case, it will fail because it won't find the key for:

```
+step.colocalisation_method_params: '{priorc1: 1e-4, priorc2: 1e-4, priorc12: 1e-5}'
```

Which is not meant to be replaced anyway.

src/ot_orchestration/utils/utils.py
javfg: It is, I agree.

Any suggestions to replace strings contained between two curly braces `{{key}}` with a `value`, without replacing those in single curly braces?

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: This is a single google batch task now, lets keep it together. I also do not think that google batch tasks will share common infrastructure if there will be any more in the future :)

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: Yeah, no need for overwrite here, it is a test artifact. For the production it's better to not overwrite.

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: Yeah, this just does not feel semantically correct.

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: The process name is `l2g_evicence`, the pipeline refers also this step in ETL evidence step, but the path does not reflect that name, rather `evidence-files/gentropy`. Lets change it for `evidence-files/l2g`.

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: It is up to you, it just feels wrong to hardcode again the project name (since you alredy had issues with this in the code comming from gentropy)

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: You never know if new datasource will not require `n1-standard-64` for instance. Let's leave it for now if you are ok with this.

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant:
```
etl_jar_base.format(version=etl_version)
```

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: It is a general rule when writing python classes
Ideally init should just set up the object properties, do not perform any computations. 
* This makes testing easier if the `read_config` and any parsers are collected into common functions.
* It aligns with OOP to move the computations to methods and perform the side effects there.
* The code is very imperative in this way, harder to follow.

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: It can stay (although similar method for gentropy exists in utils, so it is easy to find it if needed to be reused anywhere else). But definately does not need to be an instance method, you do not call `self`

src/ot_orchestration/dags/config/unified_pipeline.yaml
project-defiant: At the end we would like to have a persistant database that stores the runs and top level configurations should be included from Airflow, so we do not touch the code at all

src/ot_orchestration/dags/unified_pipeline.py
project-defiant: Yeah, I would better set the name explicitly for each run, rather then ammend some timestamp or random string to it.

src/ot_orchestration/utils/utils.py
project-defiant: There are a few possibilities, using the `string.Template` for instance or writing a custom formatter. Let's keep it this way. Just document the above example please

src/ot_orchestration/utils/utils.py
project-defiant: Use another pattern instead of `{{`

src/ot_orchestration/dags/config/gentropy.yaml
javfg: If you mean the path having `parquet` in there, it is just part of the problem. The separation between etl and gentropy is wrong too. All paths are going to be redesigned.

src/ot_orchestration/dags/config/gentropy.yaml
javfg: Agree, but I'd not change this until we review all the paths. We're just going to change it twice.

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: I do, look at line 138. ;)

src/ot_orchestration/dags/unified_pipeline.py
javfg: Let's talk this in private to avoid reply clutter.

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: I'm not 100% with you there. Can you show me where it is stated that only assigning properties in a constructor is good practice in Python?

The current design ensures the object is fully initialized and valid after construction, which is a good practice.

I think there are no side effects in the function, if there are any please point them to me. If you think about it, all it does is:

* Read config files it needs to initialize the object
* Derive values from those (like image URLs, etc)
* Compute list of steps

Regarding the code being very imperative, remember python's principle: _explicit is better than implicit_.

src/ot_orchestration/utils/utils.py
javfg: I agree with these two. I wanted to save time, but it will be better with something more legible. Should we add these to The List™? :laughing:

src/ot_orchestration/utils/utils.py
javfg: Same as above, I agree.

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: I just speak from my experience, that it is just better to have clean constructors.
>  `explicit is better than implicit` 
implict is not the same as imperative. You can still be explicit about the steps used, while contain the business logic in  declarative methods.
* read config files is a side effect, will break if the config is not there (especialy that we raise error in the read function), this will prevent you from initializing the object.
> ensuring the object is fully validated is a good practice I agree, but in this case the constructor do most of the heavy lifting of the config processing, the class can not be easily extended by addition of the new child configuration without code modification... (in summary violates  SOLID principles)

The code is logically correct and works (from your experience), we can leave it this way if you do not feel convinced to change it.

javfg: * If the config file is not there, this must break. It should break when it initializes.
* I agree reading files is a side effect, but in the strictest "pure" function meaning. The init is not changing any other parts of the application state though.
* It cannot be extended easily by adding new child configs. But I do not expect that to happen often, so I did not want to model a modular config loader class. That requires a lot more work and complexity.

project-defiant (approved):

PR #87, opened 2024-12-16
feat: ingesting FoldX mutation data from OTAR2081

### LINKED ISSUE #3676: Enable gentropy to enrich variant in-silico predictors with amino-acid variation consequences (OTAR2081)
## Context

There are a number of methods that provide variation consequences at the amino acid resolution. In such cases the method claims that changing position `n` of the sequence of protein represented by Uniprot id `u`  from amino acid `ref` to `alt` there's some predicted consequence eg. effect on protein stability. 

An example of such dataset is the [FoldX](https://home.opentargets.org/OTAR2081) project, where the team was using AlphaFold structures of all proteins to mutate all residues to all other residues and measured the resulting ddG indicating highly destabilising mutations. 

All variants in the variant index that causes amino acid change that were tested in the FoldX project needs to be annotated with the corresponding ddG values and stored in the `inSilicoPredictos` object.

## Tasks

- [x] Given there's a good chance such annotations will increase in the future, a new data model needs to be included in gentropy.
- [x] Add new datasource for FoldX.
- [x] Add logic to normalise FoldX ddG values.
- [x] Add logic to integrate dataset into the variant index.
- [ ] Add test.
- [x] Add new step to ingest data.

## Integration of FoldX dataset

- Some of the large proteins are chunked to build AlphaFold models, but the positions in the dataset are referring to the models not the full-length Uniprot sequence. The chunked Uniprot IDs need to be dropped.
- Developing logic to normalise ddG values.
- Add step to ingest data and save as static asset.

DSuveges:
## Contents

See issue for more details: [#3676](https://github.com/opentargets/issues/issues/3676)


project-defiant (commented):
A few comments (although not critical):
1. add documentation about the new dag
2. If this is a new datasource, I think we should specify a new bucket for it to store the ingestion outputs.
3. This PR does not contain the update to the `unified pipeline configuration`

DSuveges:
> 1. add documentation about the new dag
> 2. If this is a new datasource, I think we should specify a new bucket for it to store the ingestion outputs.

Good points! I'm considering create a OTAR_PPP datasource, where the data is coming from OTAR studies. Currently for PPP there's a bucket: `gs://otar013-ppp/` in the open-targets project. 

> 3. This PR does not contain the update to the `unified pipeline configuration`

What does that mean exactly? Adding the extra lines into `genetics_etl.yaml` as well?

project-defiant (approved):
Please, merge when ready

project-defiant (approved):

PR #89, opened 2025-01-06
refactor: remove gene_index from pipeline

vivienho:
Gentropy uses a `gene_index` that is generated from the target index in the platform etl via the `gene_index` step. The gentropy pipelines will now use the target index directly.

This PR removes the `gene_index` step from the orchestration pipelines.


src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -44,7 +35,7 @@ nodes:
         - gs://eqtl_catalogue_data/study_index
         - gs://ukb_ppp_eur_data/study_index
         - gs://finngen_data/r11/study_index
-      step.target_index_path: gs://ot_orchestration/releases/24.11_freeze10/gene_index
+      step.target_index_path: gs://genetics_etl_python_playground/static_assets/targets # OTP 23.12 data
```
project-defiant: I think We should use the latest target index - `gs://open-targets-pre-data-releases/24.12-uo_test-3/output/etl/parquet/targets`

src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -177,7 +168,7 @@ nodes:
       step.variant_index_path: gs://ot_orchestration/releases/24.11_freeze10/variant_index
       step.colocalisation_path: gs://ot_orchestration/releases/24.11_freeze10/colocalisation
       step.study_index_path: gs://ot_orchestration/releases/24.11_freeze10/study_index
-      step.gene_index_path: gs://ot_orchestration/releases/24.11_freeze10/gene_index
+      step.target_index_path: gs://genetics_etl_python_playground/static_assets/targets # OTP 23.12 data
```
project-defiant: As above

project-defiant (approved):
Apporved, well done!

project-defiant (commented):
See the comment, otherwise there is another PR by Daniel C that fixes the pre-commit. Wait with the merge until the https://github.com/opentargets/orchestration/pull/90 is merged

PR #92, opened 2025-01-14
fix: finngen r12 curation url updated from R11 to R12

Daniel-Considine:
Updated the curation url to use "FinnGen R12" instead of R11. 

Note: Haven't rerun ontoma EFO mapping


project-defiant (approved):

project-defiant:
@Daniel-Considine just for memory, we need to update the documentation when r12 will be rerun

PR #94, opened 2025-01-17
chore: bump dataproc image version

project-defiant:
# Context

Bump the version of the dataproc image to use pyspark 3.5.1 -> https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.2

## Implementations

1. bump dataproc image version to 2.2 (pyspark3.5) 
2. Install gentropy from github repository by ref name specified in `GENTROPY_REF` cluster metadata variable.
3. Update configuration for existing dags to match new solution
4. fix hail_home to python3.11 (used by dataproc 2.2)

The solution above allows for quicker iteration (there is no need to wait for the github actions to complete before the dags with gentropy steps can be run).

Below you can see the dataproc image 2.2 working with gentropy from `pyspark-bump` branch.

image


src/ot_orchestration/utils/dataproc.py
```diff
@@ -105,6 +105,7 @@ def create_cluster(
         num_preemptible_workers=num_preemptible_workers,
         num_workers=num_workers,
         image_version=GCP_DATAPROC_IMAGE,
+        internal_ip_only=False,
```
ireneisdoomed, line 108: Was this argument the reason why the cluster didn't have internet access?

ireneisdoomed (approved):
Thank you for updating and testing!!!
Great that it works. Only question is whether we should separate Gentropy's main entry point and cluster set up script from the Gentropy version. I know these 2 files have been virtually identical for a long time.
In general, I don't think it makes sense having to define Gentropy's locations in every single YAML. It's OK for now, but maybe we want to do this differently in the future.
Thanks anyway!

src/ot_orchestration/utils/dataproc.py
project-defiant: yes, correct!

PR #95, opened 2025-01-21
feat: sane path structure for releases

javfg:
This PR addresses the changes in the orchestrator for https://github.com/opentargets/issues/issues/3649.

It incorporates the changes in configs for PIS (https://github.com/opentargets/pis/pull/144), Ontoform (https://github.com/opentargets/ontoform/pull/15) and the ETL (https://github.com/opentargets/platform-etl-backend/pull/370).

It also solves https://github.com/opentargets/issues/issues/3652.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+  "python.analysis.typeCheckingMode": "basic",
+  "files.associations": {
+    "*.conf": "hocon",
+  },
+  "yaml.schemas": {
+    "https://raw.githubusercontent.com/opentargets/pis/refs/heads/sane-paths/.vscode/config-schema.json": "pis.yaml"
+  }
+}
--- a/src/ot_orchestration/dags/config/etl.conf
+++ b/src/ot_orchestration/dags/config/etl.conf
@@ -1,31 +1,48 @@
 // --- Templated parameters --- //
-
-common.input = "{{gcs_url}}/input"
-common.output-base_path = "{{gcs_url}}/output"
-
-etl-dag.resolve = false
-
-spark-settings.write-mode = "ignore"
-
-
-data_version = "{{release}}"
-chembl_version = "{{chembl_version}}"
-ensembl_version = "{{ensembl_version}}"
-
+common.path = "{{remote_uri}}"
 // --- End of templated part --- //
 
+common {
+  additional-outputs = []
+  output-format = "parquet"
+  metadata {
+    write-mode = "append"
+    output {
+      format = "json"
+      path = ${common.path}"/metadata/"
+    }
+  }
+}
+
+etl-dag {
+  resolve: false
+  steps = [
+    {step: "disease", dependencies: []},
+    {step: "reactome", dependencies: []},
+    {step: "expression", dependencies: []},
+    {step: "pharmacogenomics", dependencies: ["drug"]},
+    {step: "go", dependencies: []},
+    {step: "target", dependencies: ["reactome"]},
+    {step: "search_facet", dependencies: ["target", "disease", "go"]},
+    {step: "interaction", dependencies: ["target"]},
+    {step: "target_validation", dependencies: ["target"]},
+    {step: "evidence", dependencies: ["disease", "target", "drug"]},
+    {step: "association", dependencies: ["evidence", "disease"]},
+    {step: "association_otf", dependencies: ["evidence", "target", "disease", "reactome"]},
+    {step: "search", dependencies: ["target", "drug", "evidence", "association", "disease"]},
+    {step: "drug", dependencies: ["target", "disease"]},
+    {step: "literature", dependencies: ["drug", "disease", "target", "evidence"]},
+    {step: "known_drug", dependencies: ["target", "disease", "drug", "evidence"]},
+    {step: "search_ebi", dependencies: ["target", "disease", "evidence", "association"]},
+    {step: "fda", dependencies: ["drug"]},
+    {step: "otar", dependencies: ["disease"]},
+    {step: "target_engine", dependencies: ["target", "drug"]}
+  ]
+}
+
 spark-settings {
-  // Specifies the behavior when data or table already exists. Options include:
-  //   overwrite: overwrite the existing data.
-  //   append: append the data.
-  //   ignore: ignore the operation (i.e. no-op).
-  //   error or errorifexists: default option, throw an exception at runtime.
-  // Taken from `DataFrameWrite.scala` scaladoc.
+  ignore-if-exists = true  # stop step early if write-mode is ignore and files exists
   write-mode = "ignore"
-  // if set to true will short-circuit calculation of step if the output directory already exists rather than waiting
-  // until the `write` call before determining whether to proceed. This only takes effect if `spark-settings.write-mode`
-  // is set to `ignore`.
-  ignore-if-exists = true
 
   default-spark-session-config = [
     {k: "spark.driver.maxResultSize", v: "0"},
@@ -34,31 +51,15 @@ spark-settings {
   ]
 }
 
-common {
-  output-format = "parquet"
-  output-base_path = "gs://open-targets-pre-data-releases/"${data_version}"/output"
-  output = ${common.output-base_path}"/etl/"${common.output-format}
-  output = ${?OT_ETL_OUTPUT}
+# ============================= STEP CONFIGURATION =============================
+# The comment above each step shows the proper name for the step.
 
-  metadata {
-    write-mode = "append"
-    output {
-      format = "json"
-      path = ${common.output-base_path}"/metadata/"
-    }
-  }
-
-  input = "gs://open-targets-pre-data-releases/"${data_version}"/input"
-  error = ${common.output}"/errors/"
-  additional-outputs = []
-
-}
-
+# reactome
 reactome {
   inputs {
     pathways {
       format = "csv"
-      path = ${common.input}"/reactome-inputs/ReactomePathways.txt"
+      path = ${common.path}"/input/reactome/ReactomePathways.txt"
       options = [
         {k: "sep", v: "\\t"},
         {k: "header", v: "false"}
@@ -67,7 +68,7 @@ reactome {
     }
     relations {
       format = "csv"
-      path = ${common.input}"/reactome-inputs/ReactomePathwaysRelation.txt"
+      path = ${common.path}"/input/reactome/ReactomePathwaysRelation.txt"
       options = [
         {k: "sep", v: "\\t"},
         {k: "header", v: "false"}
@@ -77,28 +78,30 @@ reactome {
   }
   output = {
     format = ${common.output-format}
-    path = ${common.output}"/reactome"
+    path = ${common.path}"/output/reactome"
   }
 }
 
+# pharmacogenomics
 pharmacogenomics {
   inputs {
     pgkb {
       format = "json"
-      path = ${common.input}"/pharmacogenomics-inputs/pharmacogenomics.json.gz"
+      path = ${common.path}"/input/pharmacogenomics/pharmacogenomics.json.gz"
     }
     drug = ${drug.outputs.drug}
   }
   outputs {
     format = ${common.output-format}
-    path = ${common.output}"/pharmacogenomics"
+    path = ${common.path}"/output/pharmacogenomics"
   }
 }
 
+# expression
 expression {
   rna {
     format = "csv"
-    path = ${common.input}"/expression-inputs/baseline_expression_counts.tsv"
+    path = ${common.path}"/input/expression/baseline_expression_counts.tsv"
     options = [
       {k: "sep", v: "\\t"},
       {k: "header", v: "true"}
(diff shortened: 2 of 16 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "python.analysis.typeCheckingMode": "basic",
+  "files.associations": {
+    "*.conf": "hocon",
+  },
+  "yaml.schemas": {
+    "https://raw.githubusercontent.com/opentargets/pis/refs/heads/sane-paths/.vscode/config-schema.json": "pis.yaml"
+  }
```
DSuveges, line 8: How does this work? Would .vscode know about the schema and would help hinting the structure? I haven't seen this before.

src/ot_orchestration/dags/config/etl.conf
```diff
@@ -1,64 +1,65 @@
...
+# ============================= STEP CONFIGURATION =============================
+# The comment above each step shows the proper name for the step.
+
+# reactome
 reactome {
   inputs {
     pathways {
       format = "csv"
-      path = ${common.input}"/reactome-inputs/ReactomePathways.txt"
+      path = ${common.path}"/input/reactome/ReactomePathways.txt"
```
DSuveges, line 62: Just double checking these are the input paths for each step in the ETL, which were populated already by PIS. And the common path looks like  `gs://open-targets-pre-data-releases/partners/24.12/`?

src/ot_orchestration/dags/config/etl.conf
```diff
@@ -1,64 +1,65 @@
...
-etl-dag.resolve = false
-
-spark-settings.write-mode = "ignore"
-
-
-data_version = "{{release}}"
-chembl_version = "{{chembl_version}}"
-ensembl_version = "{{ensembl_version}}"
-
+common.path = "{{remote_uri}}"
```
DSuveges, line 2: What I find confusing with this file is that all other configuration is yaml, but this is not. I'm sure there's a good reason for this though. :D

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -45,22 +37,26 @@ steps:
...
-      step.session.output_partitions: 1
+      # INPUTS
+      step.study_index_path:
+        - gs://gwas_catalog_top_hits/study_index
+        - gs://gwas_catalog_sumstats_susie/study_index
+        - gs://gwas_catalog_sumstats_pics/study_index
+        - gs://eqtl_catalogue_data/study_index
+        - gs://ukb_ppp_eur_data/study_index
+        - gs://finngen_data/r11/study_index
+      step.target_index_path: '{release_uri}/output/target'
```
DSuveges, line 48: I know this is not actionable, but I find it inconsistent that while other components of the ETL rely on files exclusively under `'{release_uri}/input`, gentropy shops around 15 google buckets. This makes the pipelines less reproducible and kind of makes the `'{release_uri}/input` pointless. People would not be able to reproduce our pipelines anyway.

src/ot_orchestration/dags/config/etl.conf
```diff
@@ -1,64 +1,65 @@
...
-common.output-base_path = "{{gcs_url}}/output"
-
-etl-dag.resolve = false
-
-spark-settings.write-mode = "ignore"
-
-
-data_version = "{{release}}"
-chembl_version = "{{chembl_version}}"
-ensembl_version = "{{ensembl_version}}"
```
DSuveges, line 13: Where did this line go? PIS?

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -130,56 +117,92 @@ steps:
         max_run_duration: '2h'
       policy_specs:
         machine_type: n1-standard-4
-      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:{vep_version}
+      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:v{vep_version}
```
DSuveges, line 120: Isn't this `{ensembl_version}`? This can be abstracted to keep the datasets consistent across all processes.

src/ot_orchestration/utils/common.py
```diff
@@ -22,23 +20,6 @@
...
-INITIALISATION_BASE_PATH = (
-    f"gs://genetics_etl_python_playground/initialisation/{GENTROPY_VERSION}"
-)
-CONFIG_TAG = f"{INITIALISATION_BASE_PATH}/config.tar.gz"
-PACKAGE_WHEEL = (
-    f"{INITIALISATION_BASE_PATH}/gentropy-{GENTROPY_VERSION}-py3-none-any.whl"
-)
-INITIALISATION_EXECUTABLE_FILE = [
-    f"{INITIALISATION_BASE_PATH}/install_dependencies_on_cluster.sh"
-]
```
DSuveges, line 40: I don't see these paths anywhere else. Are these variables no longer used? Or these values are abstracted somehow?

DSuveges (approved):
This PR represent a systemic refactoring of pipeline configuration and many processes around orchestration. The changes all make sense, but extremely wide-spread making it very complicated to follow all potential effects. That being said, all changes make perfect sense, my questions are mostly for the purpose to me understanding what's going on.

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: No, but it should be, currently the image is bound to gentropy version, but should be bound to ensembl release. We need either add the version of the ensembl to gentropy in the CI or extract entire image.

.vscode/settings.json
javfg: Yup, you get both suggestions and syntax highlighting. It's quite nice to have.

src/ot_orchestration/dags/config/etl.conf
javfg: Yup, these are inputs for the ETL, which are outputs from PIS. If you look at [pis.yaml line 273](https://github.com/opentargets/orchestration/blob/48ced52e1aa06177cef69703798471e941134f10/src/ot_orchestration/dags/config/pis.yaml#L273) you can see that's what writing this file. (PIS paths omit the `/input` because it only works inside of it).

src/ot_orchestration/dags/config/etl.conf
javfg: It is not great, I totally agree. But the ETL uses this [HOCON format](https://github.com/lightbend/config/blob/main/HOCON.md), which seems to be popular in Scala.

As this is the actual config for the ETL it must be in that format.

src/ot_orchestration/dags/config/gentropy.yaml
javfg: It's good you point this out! I think one of the actions that remains to complete the pipeline is to make PIS entries for anything that is an input to Gentropy (just like @Tobi1kenobi did with Biosamples).

We should not forget PIS is the one in charge of putting together all we need for a release.

src/ot_orchestration/dags/config/etl.conf
javfg: And yes, the common path is the path all the way to the release name.

src/ot_orchestration/dags/config/etl.conf
javfg: We don't need those anymore.

The reason, for example for chembl, was PIS was grabbing stuff from chembl and writing it into files like:

`chembl_34_drug.json`

But now, it is just

`chembl_drug.json`

As the version is very easy to find in multiple different places, like the [config for the pipeline run itself](https://github.com/opentargets/orchestration/blob/48ced52e1aa06177cef69703798471e941134f10/src/ot_orchestration/dags/config/unified_pipeline.yaml#L10).

That way we don't have to interpolate the version here. Same for ensembl; and the data_version, which it is now in the path itself.

src/ot_orchestration/utils/common.py
javfg: I removed them because they were not used anywhere. I think @project-defiant put them there. Might want to keep it.

src/ot_orchestration/dags/config/gentropy.yaml
javfg: I'm not sure about this one, I think it's the same version as Gentropy.

Edit: I posted this one before refreshing, Szymon's reply wasn't there yet. Sorry!

.vscode/settings.json
javfg: You can check it out if you edit [pis.yaml](https://github.com/opentargets/orchestration/blob/48ced52e1aa06177cef69703798471e941134f10/src/ot_orchestration/dags/config/pis.yaml) in vscode.

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -1,40 +1,32 @@
 ---
-python_main_module: gs://genetics_etl_python_playground/initialisation/gentropy/{gentropy_version}/cli.py
+python_main_module: gs://genetics_etl_python_playground/initialisation/gentropy/v{gentropy_version}/cli.py
 
 dataproc_cluster_settings:
   cluster_metadata:
-    PACKAGE: gs://genetics_etl_python_playground/initialisation/gentropy/{gentropy_version}/gentropy-0.0.0-py3-none-any.whl
-  cluster_init_script: gs://genetics_etl_python_playground/initialisation/gentropy/{gentropy_version}/install_dependencies_on_cluster.sh
+    PACKAGE: gs://genetics_etl_python_playground/initialisation/gentropy/v{gentropy_version}/gentropy-{gentropy_version}-py3-none-any.whl
```
project-defiant, line 6: I am changing (hopefully in the next gentropy release it will be only the ref from which we want to install the package)

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -78,46 +74,37 @@ steps:
...
+        - gs://eqtl_catalogue_data/credible_set_datasets/eqtl_catalogue_susie_patched/
+        - gs://ukb_ppp_eur_data/credible_set_clean/
+        - gs://finngen_data/r11/credible_set_datasets/susie/
+      step.study_index_path: '{release_uri}/output/study'
+      # OUTPUTS
+      step.valid_study_locus_path: '{release_uri}/output/credible_set'
+      step.invalid_study_locus_path: '{release_uri}/log/credible_set_failed'
 
-  variant_conversion:
+  variant_partition:
```
project-defiant, line 90: it is sematically partial. this step converts the raw datasets to vcf format and repartitions them for the batch job. If you want to keep the names short, then mayby `collection` would be enough?

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -78,46 +74,37 @@ steps:
...
-      step.credible_set_path: '{gcs_url}/output/genetics/parquet/credible_set'
-      step.coloc_path: '{gcs_url}/output/genetics/parquet/colocalisation'
-      step.colocalisation_method: ECaviar
-      +step.session.extended_spark_conf: "{spark.sql.shuffle.partitions: '4000', spark.executor.memory: '25g', spark.sql.files.maxPartitionBytes: '25000000', spark.executor.cores: '4'}"
-      step.session.write_mode: ignore
-      step.session.output_partitions: 25
+      # OUTPUTS
+      step.output_path: '{release_uri}/intermediate/variant_partition'
 
   variant_annotation:
```
project-defiant, line 109: Could be `annotation`, would requre change in the code unfortunately

src/ot_orchestration/dags/config/unified_pipeline.py
```diff
@@ -57,7 +57,7 @@ def __init__(self) -> None:
         pis_version = settings["pis_version"]
         self.pis_config = self.init_pis_config()
         # The base image for PIS, the version tag will be appended from the config file.
-        pis_image_base = "europe-west1-docker.pkg.dev/open-targets-eu-dev/platform-input-support-test/platform-input-support-test"
+        pis_image_base = "europe-west1-docker.pkg.dev/open-targets-eu-dev/pis/pis"
```
project-defiant, line 60: why `/pis/pis` ?

src/ot_orchestration/utils/common.py
project-defiant: Nope, these are artifact, everything here is defined in the yamls, drop them

project-defiant (approved):

src/ot_orchestration/dags/config/unified_pipeline.py
javfg: I wonder about that too. I'm not sure if we're not doing it right, or GAR structure differs from the usual image registries.

The current path is: 

```
europe-west1-docker.pkg.dev/<google_cloud_project>/<repository>/<image>:<tag>
```

Not sure if there is a way we can reconcile that with `opentargets/pis`.

src/ot_orchestration/dags/config/gentropy.yaml
javfg: We've been trying to name the steps in relation to the artifact they produce (in PIS and ETL too). So even if this does some more stuff with the files, the most important part is that it generates partitioned data.

src/ot_orchestration/dags/config/unified_pipeline.py
project-defiant: It's not a deal breaker

PR #105, opened 2025-02-10
feat: ingest Loss-of-Function variant data from OTAR2075

### LINKED ISSUE #3385: Scoping integration of LoF project data into the Variant Page
The LoF project team have now shared their complete dataset with us for downstream use. 

The plan would be to add these predictions to the variant page in two places:

- [ ] variant description text 
- [ ] variant effect widget** - _currently named `in silico predictors`_ ).

Files shared with Sarah:
A csv with the curations results 
A csv with gene level curation comments
A report for this batch
An updated version of the SOPs (v4)

Below is a description of the final curation results:
"The updated final data release of 11172 pLoF variants from the gnomAD v2.1.1 database, curated by the Centre for Population Genomics (CPG). This batch includes the 10444 variants from the previous batches and 728 variants from 15 genes that were previously overlooked. Variant IDs are in the format of ‘chr-pos-ref-alt’ with separate columns for both GRCh37 and GRCh38 positions. Refer to the LoF SOP v5.pdf for a description of the flags and how they were used for curation. The code for the variant curation portal used for curation is freely available at the CPG GitHub repository at https://github.com/populationgenomics/variant-curation-portal”
 
Please note:
As we were curating TET2 we noticed that there is a note in gnomAD warning that this gene shows evidence of clonal haematopoiesis, meaning that some variants in gnomAD may be somatic variants.

vivienho:
This PR adds a new DAG to ingest LOF variant data curated in OTAR2075. The ingested output is provided as an input to the variant_index step to annotate an existing variant index.


src/ot_orchestration/dags/config/genetics_etl.yaml
```diff
@@ -150,6 +150,7 @@ nodes:
       step.variant_index_path: '{release_dir}/variant_index'
       step.amino_acid_change_annotations:
         - gs://otar013-ppp/OTAR2081_foldx/foldx_variant_annotation
+      step.lof_curation_variant_annotations_path: gs://otar013-ppp/OTAR2075_lof_curation/lof_curation_variant_annotations
```
project-defiant: Should that not be another annotation in the list above for consistency?

src/ot_orchestration/dags/config/lof_curation_ingestion.yaml
```diff
@@ -0,0 +1,20 @@
+---
+dataproc:
+  python_main_module: gs://genetics_etl_python_playground/initialisation/cli.py
+  cluster_metadata:
+    GENTROPY_REF: v2.0.1
```
project-defiant, line 5: Given that v2.1.0 is already out, I think we have to freeze the gentropy first after [your PR](https://github.com/opentargets/gentropy/pull/991) is merged before we can merge this.

project-defiant (commented):
Well done, new dag for harmonising the OTAR2075 results!

project-defiant:
@vivienho please see the comments

src/ot_orchestration/dags/config/genetics_etl.yaml
vivienho: These are different types: one is an AminoAcidVariants object and one is a VariantIndex object so should be separate inputs

src/ot_orchestration/dags/config/lof_curation_ingestion.yaml
vivienho: Yes.

src/ot_orchestration/dags/config/genetics_etl.yaml
project-defiant: I see your point, thus we already have another variant index from gnomad.

vivienho: Ah I see what you mean. I can add it to the variant_index_path instead of it being a separate parameter

vivienho:
@project-defiant I have consolidated the paths

project-defiant (approved):

PR #107, opened 2025-02-18
feat: changes for the unified pipeline run in gentropy part 

project-defiant:
# Context

1. Allow to create multiple clusters in gentropy part of unified pipeline
2. Unified pipeline configuration update to reflect changes to `genentics_etl.yaml`
3. Addition of the `l2g_training` and `pis_gold_standard` steps for the training of the L2G model.

## Notes about the coloc steps

Ability to run the colocalisation (coloc + ecaviar) steps within the gentropy part of the `unified_pipelien` requires to change the cluster configuration to allow for arbitrary number of primary workers. This is due to the fact that the cluster with `ogt_etl` autoscaling policy scales down secondary workers after `overlaps` step resulting in cache misses when the actual coloc methods are running.

To make sure the cache is only stored on the primary workers we need to use different autoscaling policy and enable the [efm mode](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/enhanced-flexibility-mode). The easiest solution for this problem with current code base it to create additional cluster **only for coloc steps** and delete it afterwards.

> [!NOTE]
> There is a possibilitity to tweak the cluster autoscaling policy to decrease the scale down factor for the secondary workers, and/or update the number of workers (the EFM mode is not feasible with neighter solution, since it has to be requrested upon cluster creation). Both solutions (although affecting a single cluster) will have to be experimented on to see the most optimal solution for the run of the coloc steps. They would also require a way to set the `DataprocUpdateClusterOperator` in a correct locations (before and afer coloc steps) to upate the cluster.

### DIFF
--- a/src/ot_orchestration/dags/config/gentropy.yaml
+++ b/src/ot_orchestration/dags/config/gentropy.yaml
@@ -210,13 +210,13 @@ steps:
       step.hf_hub_repo_id: opentargets/locus_to_gene
       step.hf_model_commit_message: 'chore: update model base model for {l2g_training} run'
       # INPUTS
-      step.credible_set_path: '{release_dir}/output/credible_set'
-      step.variant_index_path: '{release_dir}/output/variant'
-      step.feature_matrix_path: '{release_dir}/output/l2g_feature_matrix'
-      step.gold_standard_curation_path: '{release_dir}/inputs/l2g/gold_standard.json'
-      step.gene_interactions_path: '{release_dir}/output/interaction'
+      step.credible_set_path: '{release_uri}/output/credible_set'
+      step.variant_index_path: '{release_uri}/output/variant'
+      step.feature_matrix_path: '{release_uri}/output/l2g_feature_matrix'
+      step.gold_standard_curation_path: '{release_uri}/inputs/l2g/gold_standard.json'
+      step.gene_interactions_path: '{release_uri}/output/interaction'
       # OUTPUTS
-      step.model_path: '{release_dir}/etc/models/locus_to_gene_model/classifier.skops'
+      step.model_path: '{release_uri}/etc/models/locus_to_gene_model/classifier.skops'
 
   l2g_prediction:
     cluster_name: otg-etl
@@ -231,7 +231,7 @@ steps:
       # INPUTS
       step.feature_matrix_path: '{release_uri}/output/l2g_feature_matrix'
       step.credible_set_path: '{release_uri}/output/credible_set'
-      step.model_path: '{release_dir}/etc/models/locus_to_gene_model/classifier.skops'
+      step.model_path: '{release_uri}/etc/models/locus_to_gene_model/classifier.skops'
       # OUTPUTS
       step.predictions_path: '{release_uri}/output/l2g_prediction'
 
--- a/src/ot_orchestration/dags/config/gentropy.yaml
+++ b/src/ot_orchestration/dags/config/gentropy.yaml
@@ -4,14 +4,14 @@ dataproc_cluster_settings:
 
   - cluster_name: otg-etl
     cluster_metadata:
-      GENTROPY_REF: '{gentropy_version}'
+      GENTROPY_REF: 'v{gentropy_version}'
     autoscaling_policy: otg-etl
     num_workers: 2
     idle_delete_ttl: 3600 # in seconds
 
   - cluster_name: otg-etl-coloc
     cluster_metadata:
-      GENTROPY_REF: '{gentropy_version}'
+      GENTROPY_REF: 'v{gentropy_version}'
     autoscaling_policy: otg-efm
     worker_machine_type: n1-highmem-32
     num_workers: 20
@@ -129,7 +129,7 @@ steps:
         max_run_duration: '2h'
       policy_specs:
         machine_type: n1-standard-4
-      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:v{vep_version}
+      image: europe-west1-docker.pkg.dev/open-targets-genetics-dev/gentropy-app/custom_ensembl_vep:{vep_version}
       entrypoint: /bin/sh
     params:
       # INPUTS
--- a/src/ot_orchestration/dags/config/gentropy.yaml
+++ b/src/ot_orchestration/dags/config/gentropy.yaml
@@ -1,7 +1,5 @@
 ---
 
-python_main_module: gs://genetics_etl_python_playground/initialisation/cli.py
-
 dataproc_cluster_settings:
 
   - cluster_name: otg-etl
@@ -150,10 +148,9 @@ steps:
       step.vep_output_json_path: '{release_uri}/intermediate/variant_annotation'
       step.variant_annotations_path:
         - gs://gnomad_data_2/v4.1/variant_index
-        - gs://otar013-ppp/OTAR2075_lof_curation/lof_curation_variant_annotations
-      # TODO: Temporary disabled until gnetropy is updated
-      # step.amino_acid_change_annotations:
-      #   - gs://genetics_etl_python_playground/static_assets/foldx_variant_annotation
+      # - gs://otar013-ppp/OTAR2075_lof_curation/lof_curation_variant_annotations # PPP ONLY!
+      step.amino_acid_change_annotations:
+        - gs://otar013-ppp/OTAR2081_foldx/foldx_variant_annotation
       # OUTPUTS
       step.variant_index_path: '{release_uri}/output/variant'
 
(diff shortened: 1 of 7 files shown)


src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -182,30 +196,54 @@ steps:
...
+      step.hf_hub_repo_id: opentargets/locus_to_gene
+      step.hf_model_commit_message: 'chore: update model base model for {l2g_training} run'
+      # INPUTS
+      step.credible_set_path: '{release_dir}/output/credible_set'
+      step.variant_index_path: '{release_dir}/output/variant'
+      step.feature_matrix_path: '{release_dir}/output/l2g_feature_matrix'
+      step.gold_standard_curation_path: '{release_dir}/inputs/l2g/gold_standard.json'
+      step.gene_interactions_path: '{release_dir}/output/interaction'
+      # OUTPUTS
+      step.model_path: '{release_dir}/etc/models/locus_to_gene_model/classifier.skops'
```
project-defiant: @javfg I am not sure this is the correct path format, please check this one

src/ot_orchestration/dags/unified_pipeline.py
```diff
@@ -262,20 +266,37 @@ def etl_stage() -> None:
...
+                labels=labels,
+                **cluster_settings,
+            )
+            clusters[name]["delete"] = delete_cluster(
+                task_id=delete_cluster_task_id,
+                project_id=GCP_PROJECT_PLATFORM,
+                cluster_name=cluster_settings["cluster_name"],
+            )
+            clusters[name]["create_id"] = create_cluster_task_id
+            clusters[name]["delete_id"] = delete_cluster_task_id
```
javfg: This is why I keep pushing for cluster management operators. You are extracting a big chunk of their responsibilities here; which will make the orchestrator much more confusing.

What about something like this? (This changes the whole gentropy part, lines 228:333).

```
    if len(config.gentropy_step_list):
        clusters = {}

        def _get_create_cluster(cluster_name: str):
            if not clusters.get(cluster_name):
                clusters[cluster_name]["create"] = create_cluster(
                    cluster_name=create_cluster_name(cluster_name),
                    project_id=GCP_PROJECT_PLATFORM,
                    **config.gentropy_dataproc_cluster_settings[cluster_name],
                    idle_delete_ttl=90 * 60,
                )
            return clusters[cluster_name]["create"]

        def _get_delete_cluster(cluster_name: str):
            if not clusters.get(cluster_name):
                clusters[cluster_name]["delete"] = delete_cluster(
                    cluster_name,
                    project_id=GCP_PROJECT_PLATFORM,
                )
            return clusters[cluster_name]["delete"]

        @task_group(group_id="gentropy_stage")
        def gentropy_stage() -> None:
            for step_name in config.gentropy_step_list:
                step_config = config.gentropy_step(step_name)
                labels = StepLabels("gentropy", step_name, config.is_ppp)

                match step_name:
                    case "gentropy_variant_annotation":
                        r = VepAnnotateOperator(
                            job_name=create_name("variant_annotation"),
                            task_id=f"run_{step_name}",
                            project_id=GCP_PROJECT_PLATFORM,
                            **step_config["params"],
                            google_batch=step_config["google_batch"],
                            labels=labels,
                        )
                        chain(r)
                    case _:
                        r = submit_gentropy_step(
                            cluster_name=config.gentropy_dataproc_cluster_settings.get(step_name, "default"),
                            step_name=step_name,
                            project_id=GCP_PROJECT_PLATFORM,
                            params=step_config["params"],
                            labels=labels,
                        )
                        chain(_get_create_cluster(step_name), r, _get_delete_cluster(step_name))

        r = gentropy_stage()
```

At least this way we reduce the complexity into those two functions. You should test it, I haven't run the code and it might require some changes.

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -1,17 +1,25 @@
...
+    num_workers: 2
+    idle_delete_ttl: 3600 # in seconds
+
+  - cluster_name: otg-etl-coloc
+    cluster_metadata:
+      GENTROPY_REF: 'v{gentropy_version}'
+    autoscaling_policy: otg-efm
+    worker_machine_type: n1-highmem-32
+    num_workers: 20
+    allow_efm: true
```
javfg, line 18: Better to have a dict here, so we can reference keys.

```suggestion
dataproc_cluster_settings:
  default:
    cluster_name: otg-etl
    cluster_metadata:
      GENTROPY_REF: 'v{gentropy_version}'
    autoscaling_policy: otg-etl
    num_workers: 2
    idle_delete_ttl: 3600 # in seconds

  colocalisation_coloc:
    cluster_metadata:
      GENTROPY_REF: 'v{gentropy_version}'
    autoscaling_policy: otg-efm
    worker_machine_type: n1-highmem-32
    num_workers: 20
    allow_efm: true
```

src/ot_orchestration/dags/config/gentropy.yaml
javfg:
```suggestion
      step.credible_set_path: '{release_uri}/output/credible_set'
      step.variant_index_path: '{release_uri}/output/variant'
      step.feature_matrix_path: '{release_uri}/output/l2g_feature_matrix'
      step.gold_standard_curation_path: '{release_uri}/inputs/l2g/gold_standard.json'
      step.gene_interactions_path: '{release_uri}/output/interaction'
      # OUTPUTS
      step.model_path: '{release_uri}/etc/model/locus_to_gene_model/classifier.skops'
```

src/ot_orchestration/dags/config/gentropy.yaml
```diff
@@ -182,30 +196,54 @@ steps:
...
       step.run_mode: predict
       step.l2g_threshold: 0.05
-      # INPUTS
       step.download_from_hub: false
       step.hf_hub_repo_id: opentargets/locus_to_gene
+      # INPUTS
       step.feature_matrix_path: '{release_uri}/output/l2g_feature_matrix'
       step.credible_set_path: '{release_uri}/output/credible_set'
-      step.model_path: '{release_uri}/input/l2g_prediction/classifier.skops'
+      step.model_path: '{release_dir}/etc/models/locus_to_gene_model/classifier.skops'
```
javfg:
```suggestion
      step.model_path: '{release_uri}/etc/model/locus_to_gene_model/classifier.skops'
```

src/ot_orchestration/dags/config/pis.yaml
```diff
@@ -211,6 +211,11 @@ steps:
       source: http://purl.obolibrary.org/obo/go.obo
       destination: go/go.obo
 
+  gold_standard:
```
javfg:
```suggestion
  l2g_gold_standard:
```

Although if you think we might need more files downloaded for gentropy, I'd name the step `gentropy`, unless the amount becomes worthy of splitting.

src/ot_orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -42,6 +43,7 @@ steps:
     ppp_only: true
   pis_expression:
   pis_go:
+  pis_gold_standard:
```
javfg:
```suggestion
  pis_l2g_gold_standard:
```

src/ot_orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -197,10 +199,18 @@ steps:
       - gentropy_colocalisation_ecaviar
       - gentropy_variant
 
-  gentropy_l2g_prediction:
+  gentropy_l2g_training:
     depends_on:
+      - gentropy_credible_set
+      - gentropy_variant
+      - gentropy_l2g_feature_matrix
+      - pis_gold_standard
```
javfg:
```suggestion
      - pis_l2g_gold_standard
```

javfg (changes requested):

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: I disagree, the names should be defined within the dictionary, this will allow for creating the config model for the operator later.

src/ot_orchestration/dags/config/gentropy.yaml
project-defiant: good catch!

src/ot_orchestration/dags/unified_pipeline.py
project-defiant: I have changed the logic (hopefully it is more manageable now)

javfg (approved):

javfg:
I like it, let's merge quickly and check it out. There are a couple of typos in the docstrings but I'll fix them in my subsequent PR.

PR #111, opened 2025-02-28
feat: manual build for gentropy based image

project-defiant:
# Context

In order to test run the google batch job I need to be able to tweak the build for docker image for the current version (2.2.0.rc-5) of gentropy. For futureproof I want to be able to run the process manually.


d0choa (approved):
`workflow_dispatch:` empty is ok right?

project-defiant:
> `workflow_dispatch:` empty is ok right?

Yes it does not need any other parameter -> see [docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch)

PR #124, opened 2025-04-11
chore: added docs describing interval dataset locations

project-defiant:
# Context

We want to have a separate bucket that will host all interval datasets to make sure we keep track to it and have documentation describing the inputs.


xyg123 (approved):
Looks good!

PR #131, opened 2025-06-06
chore: allow specifying ADC from makefile

project-defiant:
This PR includes:

1. Allow user to specify the google credentials from a custom file when spinning the local instance of airflow service
2. Readme update to reflect the change.


DSuveges (approved):
The branch works as expected. I could reach both the local and remote airflow instance.

PR #134, opened 2025-06-16
feat(ppp): per step config overrides

project-defiant:
# Context

In order to allow the user to run the Unified Pipeline in PPP configuration we require to update the in-step parameters for 3 gentropy steps:
- study
- variant
- credible_set

To allow for this functionality I have implemented an override strategy on the existing configs that uses dictionary DeepMerge on the config `steps` dict.

The overrides does two main things:
1. Merge the default dict config within the `steps` key with override `steps` (deep merge) if the type is dict
2. Override the default list config within the `steps` key with override `steps` if the type is list

This is a PoC of the solution, as the context may grow bigger in time and may require creating the recursive parser with custom rules.


javfg (approved):
Looks good!

PR #136, opened 2025-07-02
chore: set local instance as default in make

javfg:
This makes more sense, as the cloud instance is usually reserved for pipeline runs and will be used more sparingly. Also, by defaults users should not accidentally spawn a cloud instance.


javfg:
I'm putting this in the `dev-runs` PR.

PR #142, opened 2025-08-20
feat: add make rule to wipe airflow vm

ireneisdoomed:
This adds a new rule to the Makefile to delete the Airflow instance on Google Cloud.
I've tested it works when the VM was up and running, and that it didn't crash when there's no VM on.


ireneisdoomed:
@SzymonSzyszkowski Feel free to review, if you want to

project-defiant (approved):
👍🏼 it may be now more clean how the VM gets deleted :D

PR #149, opened 2025-11-03
feat(staging): 25.12 genetics staging data update

project-defiant:
# Context

This PR adds the configuration changes to the gentropy staging dags used to generate 25.12 release data.

It also unifies the way to build orchestration images. The updated logic is in the `artifact.yaml` file that corresponds to the github action responsible for building and pushing the orchestration related Docker images.

## Implementations

The PR includes:

1. Change in how the `artifact.yaml` is triggered - now only manually.
2. If `artifact.yaml` is triggered from the branch, no image is build, if it is triggered from tag, then images are build and pushed to the registry - this ensures no overwrites (branch name is not used as image tag).
3. Push images to GAR on `open-targets-genetics-dev` project via OIDC, not using **long lifetime  service account** anymore.
4. Push images to GitHub container registry
5. Build attestations and include in GAR
6. Short docs about images.
7. Adds new harmonisation and fine-mapping config changes for 25.12 release

The most important change is that now we build both images:
- gentropy overwrite image used for running gwas catalog harmonisation batch job
- orchestration service image
and include them into the GitHub Container Registry for public usage.

Action was tested on `v25.12-staging.5` which produced artifacts for the repository and GAR.


src/orchestration/dags/config/gwas_catalog_sumstats_susie_finemapping.yaml
project-defiant: I need to update this

ireneisdoomed: Is this still pending? Should it be addressed in a different PR? @project-defiant

project-defiant: It's ready now. I had to update the version.

docs/containerization/README.md
ireneisdoomed, line 1: Typo: `containerization`

Makefile
```diff
@@ -96,17 +96,3 @@ upload-finngen-meta-readme: ## upload finngen-meta readme to the bucket
 	@gcloud storage rsync docs/datasources/finngen_meta_data gs://finngen_ukb_mvp_meta_data/docs
 
 update-bucket-docs: upload-eqtl-catalogue-bucket-readme upload-ukb-ppp-bucket-readme upload-finngen-bucket-readme upload-gwas-catalog-buckets-readme upload-gnomad-bucket-readme upload-intervals-bucket-readme upload-finngen-meta-readme ## upload readmes to the datasource buckets
-
-build-gentropy-gcs-image: ## build image that overwrited gentropy with tools specific for orchestration and google cloud
```
ireneisdoomed, line 100: You are deleting this because you're leaving the image creation for CI/CD?

ireneisdoomed (approved):
Great work, thanks!!! As far as the Gentropy config goes, everything makes sense. I'd suggest that someone else does the preprocessing part for next run.

In terms of the image generation, I like the fact that tag creation triggers the image builds, instead of doing it manually. Do you have an opinion against it?

One q: Can you explain why we need another Gentropy image? We have different configuration between Gentropy and the orchestration, but what is the difference to justify a new image? I guess it's not Gentropy parameters. Thanks

Only change that i'd request is the typo in the containerization docs

Makefile
project-defiant: Yes, this is no longer needed.

project-defiant:
> Great work, thanks!!! As far as the Gentropy config goes, everything makes sense. I'd suggest that someone else does the preprocessing part for next run.

I agree

> In terms of the image generation, I like the fact that tag creation triggers the image builds, instead of doing it manually. Do you have an opinion against it?

 Yes and no, the fact that we host images in GC and GitHub (makes it easier to access by community) means that we would release new image every time for community when new tag is in the place which may be too often.

> One q: Can you explain why we need another Gentropy image? We have different configuration between Gentropy and the orchestration, but what is the difference to justify a new image? I guess it's not Gentropy parameters. Thanks

For two reasons:
1) gentropy image (from orchestration) is used for harmonisation and qc. This is done currently via a script that needs google sdk to download and upload the harmonised files. I assumed that gentropy should be as much cloud agnostic as possible, hence the default gentropy image (from gentropy) does not have google cloud sdk, but the orchestration one does.

> Only change that i'd request is the typo in the containerisation docs
Not sure what the typo is TBH 😕

docs/containerization/README.md
project-defiant: ??

ireneisdoomed:
Thank you for the explanations, I agree and it all makes sense

docs/containerization/README.md
project-defiant: renamed directory, thanks!

PR #152, opened 2025-11-06
Run PTS steps in dataproc when there are pyspark tasks

javfg:
This PR enables running PTS steps in a dataproc cluster when there are pyspark tasks in them.

It also introduces a stub of `Step` class, where we can slowly refactor config and settings that belong to a step. I only implemented what I needed for the main PR goal.

Most of the lines in additions/deletions are updating the configs for PIS/PTS/ETL, so don't be scared. :smile:

### DIFF
--- a/src/orchestration/assets/dataproc_pts_run.py
+++ b/src/orchestration/assets/dataproc_pts_run.py
@@ -1,0 +1,12 @@
+#!/opt/conda/default/bin/python
+# -*- coding: utf-8 -*-
+import sys
+
+from pts.core import main
+
+if __name__ == "__main__":
+    if sys.argv[0].endswith("-script.pyw"):
+        sys.argv[0] = sys.argv[0][:-11]
+    elif sys.argv[0].endswith(".exe"):
+        sys.argv[0] = sys.argv[0][:-4]
+    sys.exit(main())
--- a/src/orchestration/dags/config/pis.yaml
+++ b/src/orchestration/dags/config/pis.yaml
@@ -1,15 +1,19 @@
 ---
 work_path: /mnt/disks/work
 log_level: INFO
-pool: 16
+pool_size: 16
 
 release_uri: '{{release_uri}}'
 scratchpad:
   chembl_version: '{{chembl_version}}'
   efo_version: '{{efo_version}}'
   ensembl_version: '{{ensembl_version}}'
-  hpo_version: '2025-01-16'
-  mondo_version: '2025-02-04'
+  gencode_version: '{{gencode_version}}'
+  depmap_version: '{{depmap_version}}'
+  hpo_version: '{{hpo_version}}'
+  mondo_version: '{{mondo_version}}'
+  ot_curation: '{{ot_curation}}'
+  probes_drugs_version: '{{probes_drugs_version}}'
 
 steps:
   #: BIOSAMPLE STEP :###############################################################################
@@ -50,14 +54,9 @@ steps:
 
   #: DRUG STEP :####################################################################################
   drug:
-    - name: find_latest chemical probes
-      source: gs://otar001-core/ChemicalProbes/annotation
-      scratchpad_key: latest_chemical_probes
     - name: copy chemical probes
-      requires:
-        - find_latest chemical probes
-      source: ${latest_chemical_probes}
-      destination: input/drug/chemicalprobes.json.gz
+      source: https://www.probes-drugs.org/media/download/probes/pd_export_${probes_drugs_version}_probes_standardized.xlsx
+      destination: input/target/chemicalprobes/probes.xlsx
 
     - name: find_latest drugbank annotation
       source: gs://otar001-core/DrugBank/annotation
@@ -78,7 +77,6 @@ steps:
         - efo_id
         - max_phase_for_ind
         - indication_refs
-
     - name: elasticsearch chembl drug warning
       url: https://www.ebi.ac.uk/chembl/elk/es
       destination: input/drug/chembl_drug_warning.jsonl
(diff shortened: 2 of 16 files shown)


src/orchestration/assets/dataproc_pts_run.py
```diff
@@ -0,0 +1,12 @@
...
+# -*- coding: utf-8 -*-
+import sys
+
+from pts.core import main
+
+if __name__ == "__main__":
+    if sys.argv[0].endswith("-script.pyw"):
+        sys.argv[0] = sys.argv[0][:-11]
+    elif sys.argv[0].endswith(".exe"):
+        sys.argv[0] = sys.argv[0][:-4]
```
project-defiant, line 11: You may have to explain to me why we need the executable to drop the extension. Are you thinkig someone might attempt to run this from windows ?

src/orchestration/dags/config/pis.yaml
```diff
@@ -1,15 +1,18 @@
...
 log_level: INFO
-pool: 16
+pool_size: 16
 
 release_uri: '{{release_uri}}'
 scratchpad:
   chembl_version: '{{chembl_version}}'
   efo_version: '{{efo_version}}'
   ensembl_version: '{{ensembl_version}}'
+  depmap_version: '2025Q2'
```
project-defiant: This may be an oversight from my side, but I need some explanation.
Does the templating works as follows:
1. Take the ensembl_version from the up config and substitute when `{{ensembl_version}}` is in PIS
2. Take the scratchpad ensembl_version and substitute in all fields in the PIS config below where we find `${ensembl_version}`

???

If that is correct, we need to update the `gencode_version` to be included in the scratchpad as well.

Besides that I would keep all provider versions in one place (hopefully unified pipeline config, not here)

src/orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -1,27 +1,27 @@
...
+run_name: 'jferrer/test-edp-1'
 
 # Whether this will be a partner preview platform release (PPP)
-is_ppp: true
+is_ppp: false
 
 # Versions for software and data
-pis_version: '25.2.1'
-pts_version: '25.2.1-rc.1'
+pis_version: '25.2.3.dev.18'
```
project-defiant, line 15: I think you are losing some of the changes from dev. The vep_version should be 115.2 and so on.

src/orchestration/models/pts_step.py
```diff
@@ -0,0 +1,133 @@
...
+        PTSStep: The created PTSStep instance.
+    """
+    short_name = name.split("_", 1)[1]
+    step_tasks = config.pts.config.get("steps", {}).get(short_name, {})
+    if len(step_tasks) == 0:
+        raise ValueError(f"no tasks found for pts step {name}")
+
+    # if any task inside a step is a pyspark task, then this is a dataproc step
+    for task in step_tasks:
+        if task.get("name", "").startswith("pyspark"):
```
project-defiant, line 35: NIT: It may be by desing, but I feel that adding more config to the step name (like distinction between pyspark vs gce task) may give us a headache in the future. Arguably field `infra` would make sense here.

Either way I hope it's well documented somewhere. Not a breaker.

src/orchestration/models/pts_step.py
```diff
@@ -0,0 +1,133 @@
...
+        config: UnifiedPipelineConfig,
+    ):
+        """Initialize a PTSDataprocStep instance.
+
+        Args:
+            name (str): The name of the step.
+            config (AppConfig): The config object for PTS.
+        """
+        super().__init__(name, config)
+        self.runs_on_cluster = True
```
project-defiant, line 99: I guess it does not make sense to run dataproc step outside the cluster and you can always make the PTS pyspark based step to run via GCE with not including `pyspark` in the name of the step.

project-defiant (approved):
Clean, have a look at the comments. I see just one dealbreaker - missing changes from dev. Otherwise approved!

Fantastic job!

src/orchestration/assets/dataproc_pts_run.py
javfg: I just copied the _binary_ uv generates in the `venv/bin/` folder when you install PTS. :smile:

src/orchestration/dags/config/pis.yaml
javfg: The way you described is correct.

And you are right, those need to be extracted into the `unified_pipeline.yaml` file. I just haven't done it because they are new and I did not care about them for this PR in particular.

src/orchestration/dags/config/unified_pipeline.yaml
javfg: I'll bring back the missing changes!

src/orchestration/models/pts_step.py
javfg: You are right, I would not add anything more to the step name. However, this is looking into the PTS config itself, not the airflow step name. If there is a pyspark otter task, then this must run on dataproc.

src/orchestration/models/pts_step.py
javfg: This ties with my reply above. We need to specify somehow that this runs in a cluster, as the step name only includes the app (`pts_`, `pis_`).

I agree this bool is not the best way. I actually thought about removing it before opening the PR and figure out another way. But we can change it afterwards easily I think, given we're in a hurry.

project-defiant: agree to agree :D

javfg:
I rebased onto `dev` again and also added parameters for those other PIS sources.

javfg:
Finally passing tests :sweat_smile:

project-defiant:
@javfg I leave merge up to you

PR #155, opened 2025-11-14
feat(l2g predictions) batch job

project-defiant:
# Context

The dataproc job for `l2g_predictions` with current setup is very long (~5h) and utilizes cluster only for io operations. The shap explanations are run on a single machine. 

To address the time needed to compute shap, we are bringing back the google batch job version of the step.

## Implementations

See https://github.com/opentargets/orchestration/pull/110/files#diff-4ff77775e9b0896f7036d224cf5b0c2b582ad0caa56ad9e701c38bd0fd340594R75-R76

- [x] addition of `$OUTPUT_PARTITION` into the batch task, as spark does not allow for async append write, that can cause issues requiring to retrigger the job.


ireneisdoomed (approved):
👩‍🚀 Let's try
Thanks!

PR #156, opened 2025-11-19
feat: prettify unified pipeline graph in Airflow

d0choa:
**WARNING: Don't merge until the iteration of the pipeline run is finished due to Airflow database losses**

At the moment, the unified pipeline graph is hard to follow because the task groups impose a grouping by PIS, ETL, PTS, and Gentropy. Because now many paths go from one codebase to another and back is hard to read in the graph view of Airflow.

This PR removes these high-level task groups, which only had aesthetic implications.

Together with @ireneisdoomed  we checked the effects of these changes, and the view is very nicely improved, making it easier to navigate the graph of dependencies. However, this change alters the Airflow database, altering the task identifier and forgetting about previous runs.

### DIFF
--- a/src/orchestration/dags/unified_pipeline.py
+++ b/src/orchestration/dags/unified_pipeline.py
@@ -76,7 +76,6 @@ with DAG(
     #   d. Diff   — the state and decide the step does not need to run.
     #   e. End    — the step (does nothing).
     # ==============================================================================================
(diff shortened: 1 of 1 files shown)


javfg:
It certainly looks better :smile: 
image

javfg (approved):

PR #157, opened 2025-11-26
fix: re-wiring variant index generation

### LINKED ISSUE #4157: Fixing variant identifier generation
## Context

Long indels caused a headache, when we tried to use them as identifiers for variants. We have decided to generate hashes based on chromosome position and alleles. We used this hashed identifier in evidence and the variant details were resolved through variant index.

To build the variant index, evidence input was collected from all relevant sources prior ETL evidence step and a hashing step was done in both evidence and variant index step. This process has two flaws:
- It was a risk that the two processes might yield different hashes
- The variant index is based on un-validated evidence causing variants without any phenotypic information (#3894) 

Upon exploding the evidence step allowed us to build evidence index based on variants coming from valid evidence, however this also caused an unforeseen problem: as the hashed identifier replaced the original identifier, the hash could not be used to build the index due to data loss. 

## Proposed solution

- <s>**PTS evidence step**: upon hashing the variant identifier, we need to retain the original variant as a new column called `variantLabel`. This will prevent information loss.</s>
- <s>**Gentropy variant_partition step**: upon collecting variant identifiers from evidence files, this new column should be used instead of the current `variantId` because that might contain hashes.</s>

just reverting the orchestration configuration that the variant index generation relies on the unprocessed json evidence input.

DSuveges:
## Context

If variant index generation picks up post-processed evidence, the already hashed variant identifiers are lost. This PR revert back to the original process, when the VEP input was generated based on the JSON evidence input.

This also means the dependencies of this step changes: EVA and Uniprot evidence processing is independent from variant index generation.

### DIFF
--- a/src/orchestration/dags/config/unified_pipeline.yaml
+++ b/src/orchestration/dags/config/unified_pipeline.yaml
@@ -408,11 +408,10 @@ steps:
       - gentropy_credible_set
   gentropy_variant_partition:
     depends_on:
-      - pts_evidence_postprocess_uniprot_variants
-      - pts_evidence_postprocess_clinvar
-      - pts_evidence_postprocess_clinvar_somatic
       - etl_pharmacogenomics
       - gentropy_credible_set
+      - pis_evidence_uniprot_variants
+      - pis_evidence_eva
   gentropy_variant_annotation:
     depends_on:
       - gentropy_variant_partition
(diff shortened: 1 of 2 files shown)


src/orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -408,9 +408,6 @@ steps:
       - gentropy_credible_set
   gentropy_variant_partition:
     depends_on:
-      - pts_evidence_postprocess_uniprot_variants
```
d0choa, line 411: I think we need this here:
- pis_evidence_eva
- pis_evidence_uniprot_variants

DSuveges: Added.

ireneisdoomed (approved):
This is good, thanks!

PR #161, opened 2026-01-19
feat: add interval features to l2g

project-defiant:
# Context

For the upcoming March release we want to add enhancer2Gene fetures to L2G. This PR links the new gentropy version that brings the functionality and adds the `enhancerToGene` dataset and step to L2G feature matrix step as dependencies.

The QC of the intervals will be done in next PRs.

### DIFF
--- a/src/orchestration/dags/config/gentropy.yaml
+++ b/src/orchestration/dags/config/gentropy.yaml
@@ -175,7 +175,7 @@ steps:
       step.credible_set_path: '{{release_uri}}/output/credible_set'
       step.feature_matrix_path: '{{release_uri}}/intermediate/l2g_feature_matrix'
       step.gold_standard_curation_path: '{{release_uri}}/input/l2g/gold_standard.json'
-      step.intervals_path: '{{release_uri}}/output/enhancerToGene'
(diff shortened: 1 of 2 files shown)


src/orchestration/dags/config/gentropy.yaml
```diff
@@ -175,6 +175,7 @@ steps:
       step.credible_set_path: '{{release_uri}}/output/credible_set'
       step.feature_matrix_path: '{{release_uri}}/intermediate/l2g_feature_matrix'
       step.gold_standard_curation_path: '{{release_uri}}/input/l2g/gold_standard.json'
+      step.intervals_path: '{{release_uri}}/output/enhancer_to_gene'
```
xyg123: Also needed in the feature matrix config on line 160

xyg123 (approved):
Thank you!  Just needs the step.intervals_path added to the feature matrix config.  Also, this is the "slim" version of e2g, where only the maximum score between each interval to gene is kept.

src/orchestration/dags/config/gentropy.yaml
project-defiant: Thanks, it looks like the `intervals_path` is only required for feature matrix step.

PR #163, opened 2026-01-28
feat: add pis_encode and gentropy_enhancer_to_gene steps

### LINKED ISSUE #3954: rE2G dataset sync
As a **developer** I want **to move the code used to ingest the rE2G datasets to PIS** because **I want to make sure the process is reproducible.**

## Background

The code used to ingest the rE2G datasets from ENCODE website is provided in https://github.com/project-defiant/encode-crawler/tree/main/src/crawler. The main task is to download 1460 small bed files in concurrent manner and upload them to the google cloud storage. The good idea is to move the codebase to the PIS repo so we maintain the reproducibility.

## Tasks

- [ ] Add tasks to download rE2G datasets to PIS as a separate step
- [ ] Define how the input query file is downloaded using ENCODE api and use it as a manifest for the download
## Acceptance tests

How do we know the task is complete? 

1. Task is fully functional in Orchestration

project-defiant:
# Contents

This PR adds the `pis_encode`  and  `gentropy_enhancer_to_gene` steps to the Unified Pipeline.

## Comparison for running the `gentropy_enhancer_to_gene`

Original intervals (25.12): 48,854,316                                                        
Output intervals: 48,810,405
Excluded intervals: 68,154,612

```
|[Interval has a duplicate with different score]                                                                                                          |42112   |
|[Target/gene identifier could not match to reference]                                                                                                    |129680  |
|[Biosample identifier was not found in the reference, Target/gene identifier could not match to reference]                                               |1798    |
|[Biosample identifier was not found in the reference, Interval has a duplicate with different score, Score was above or below specified thresholds]      |672     |
|[Score was above or below specified thresholds, Target/gene identifier could not match to reference]                                                     |234456  |
|[Biosample identifier was not found in the reference, Score was above or below specified thresholds, Target/gene identifier could not match to reference]|3038    |
|[Biosample identifier was not found in the reference, Interval has a duplicate with different score]                                                     |712     |
|[Interval has a duplicate with different score, Score was above or below specified thresholds]                                                           |55758   |
|[Score was above or below specified thresholds]                                                                                                          |66180554|
|[Biosample identifier was not found in the reference, Score was above or below specified thresholds]                                                     |868917  |
|[Biosample identifier was not found in the reference]    
```
Validation failures were due to:
- biosample not found in biosampleIndex
- geneId not found in targetIndex
- when two same intervals have diverging scores for the same geneId, biosampleId mapping. 
- exclusion of intervals with scores below 0.6

## Execution time

- `pis_encode` -> 5min
- `gentropy_enhancer_to_gene` -> 15 min

## Architecture

The Interval ingestion logic:
- `pis_encode` 
    (1) Download the manifest via pis from ENCODE using the query url
    (2) Download the biosample mapping from ot curation repo
    (3) Download the chromosome index (used for checking the interval bounds are within expected range)
    (4) Based on manifest, download specific ENCODE bed files with the intervals
- `gentropy_enhancer_to_gene`
    (1) Parse the bed files to `Intervals` dataset and run the validation based on `target`, `biosample` and `chromosome` datasets.

### DIFF
--- a/src/orchestration/dags/config/unified_pipeline.yaml
+++ b/src/orchestration/dags/config/unified_pipeline.yaml
@@ -1,21 +1,21 @@
 ---
-release_name: '25.12-ppp'
+release_name: '26.01-interval-qc-2'
 
 # If `is_dev` is true:
 #   * the bucket used for the run will be `opentargets-pipeline-runs`
 #   * the prefix inside the bucket will be `run_name`
 #         which defaults to a timestamp if not provided.
 is_dev: true
-run_name: 'szsz/25.12-ppp-testrun-3'
+run_name: 'szsz/testrun-interval-qc-2'
 
 # Whether this will be a partner preview platform release (PPP)
 is_ppp: true
 
 # Versions for software and data
-pis_version: '25.12.0'
+pis_version: '26.1.0-dev.1'
 pts_version: '25.12.1'
 etl_version: '25.12.1'
-gentropy_version: '3.2.0-rc.1'
+gentropy_version: '3.2.0-dev.6'
 vep_version: 'release_115.2'
 chembl_version: '36'
 efo_version: '3.83.0'
@@ -49,6 +49,7 @@ gnomad_version: '4.1'
 #
 steps:
   # PIS STEPS
+  pis_encode:
   pis_biosample:
   pis_disease:
   pis_drug:
@@ -436,6 +437,11 @@ steps:
   gentropy_biosample:
     depends_on:
       - pis_biosample
+  gentropy_enhancer_to_gene:
+    depends_on:
+      - pis_encode
+      - etl_target
+      - gentropy_biosample
   gentropy_study:
     depends_on:
       - pts_target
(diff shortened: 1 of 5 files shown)


src/orchestration/dags/config/unified_pipeline.yaml
```diff
@@ -1,21 +1,21 @@
 ---
-release_name: '25.12-ppp'
+release_name: '26.01-interval-qc-2'
```
xyg123, line 2: Is this OK to leave like this? Will it be updated the next run?

xyg123 (approved):
This PR implements steps to fetch raw interval datasets from ENCODE, and to run it through the QC and filtering step implemented in gentropy.  Very nice.

src/orchestration/dags/config/unified_pipeline.yaml
project-defiant: I think we can leave it (this will respect the config for the test run). This shall be updated before next release run.

PR #166, opened 2026-02-13
feat: unify the retry policy for batch jobs

project-defiant:
# Context

Batch jobs are prone to mutiple failures explained in https://docs.cloud.google.com/batch/docs/troubleshooting#reserved-exit-codes. This PR attempts to unify the way how we handle the batch native exit codes, so the subsequent jobs are finished properly.

This was due to the issues I had while running L2G predictions -> ex. https://console.cloud.google.com/batch/jobsDetail/regions/europe-west1/jobs/up-l2g-prediction-ba832-job-1-20260211-191214/details?project=open-targets-eu-dev

### DIFF
--- a/src/orchestration/dags/config/gentropy.yaml
+++ b/src/orchestration/dags/config/gentropy.yaml
@@ -215,7 +215,7 @@ steps:
         memory_mib: 7000
         boot_disk_mib: 4000
       task_specs:
-        max_retry_count: 0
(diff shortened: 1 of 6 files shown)


DSuveges (approved):
Refactoring retry policy for batch job by making retry managed not by the specific job operator (vep or finemapping) but the a more general batch utility function `create_task_spec`.

PR #171, opened 2026-03-19
feat(ppp): 26.03-ppp release update

project-defiant:
# Context

This PR encompass all changes imlemented for 26.03-ppp release.

## Implementations

1. Expose provisioning mode to google batch jobs (change lost from public run)
2. Enhance colocalisation cluster setup (still some issues, but no failures due to OOM)
3. Add cluster for `pts_association` tasks -> speeds up associations 3x
4. Add `pis_impc` task as dependency of `pts_mouse_phenotype` (missing dependency)
5. use STANDARD instead of SPOT provisioning_mode for vep annotations and l2g_predictions batch jobs to ensure tasks do not fail
6. Allow for `secondary_worker_disk_size`, `secondary_worker_machine_type`, `secondary_worker_disk_type` options for dataproc operator (currently secondary workers inherit the options from primary workers. This is important for the [`efm_mode`](https://docs.cloud.google.com/dataproc/docs/concepts/configuring-clusters/enhanced-flexibility-mode) which allows us to store all shuffle partitions in STANDARD machines, while retaining computation on cheaper SPOT machines). Without this change we pay for the disks attached to secondary workers, which we do not use at all.)
7. Added PIS tasks to move gentropy inputs (molqtl, molqtl_ppp, gwas, gwas_ppp) to release inputs and refactored gentropy jobs to use the inputs instead of staging buckets.
8. Added pydantic model for ClusterConfig and pydantic as dependency


DSuveges (approved):
Thank you, all changes make sense!

PR #199, opened 2026-05-26
feat(25.06) staging update

project-defiant:
# Context

This PR aligns the configurations used to produce the staging buckets and updates changelogs for individual buckets.


DSuveges (approved):
All makes sense

PR #204, opened 2026-06-01
fix(dag bag) tests

project-defiant:
# Context

This PR fixes the test path to the dag_bag. Currently tests referencing dag_bag fixture does not work properly, as the path they reflect do not contain any dags.


ireneisdoomed (approved):
Thank you Szymon.
This is to solve the problem where a recent commit had a breaking change to the unified pipeline DAG that wasn't picked up by the tests

PR #195, opened 2026-05-18
feat(literature): adopt validated pts_literature pipeline, retire etl_literature

d0choa:
## Summary

Replaces the legacy single-step `etl_literature` with the 6-step
`pts_literature_*` graph validated in PTS runs 014-016 at full-EPMC
scale (~730 M mention rows). Introduces a shared `pts_literature`
Dataproc cluster (4 base workers + autoscaling up to 50 non-preemptible
secondaries) that all 6 steps land on via longest-prefix match.

### Pipeline graph

```
                    pis_literature
                          │
pis_ontoma  pts_disease  pts_target  pts_drug_molecule
    │           │             │              │
    └───────────┴──────┬──────┴──────────────┘
                       │
   pts_literature_ontoma_lut_generation
                       │
       pts_literature_publication_match
       ┌───────────────┼───────────────┐
       │               │               │
 entity_lut        embedding      cooccurrence_evidence
                       │               │
                  vector         (feeds pts_evidence_
                                  postprocess_europepmc)
```

### Changes

- `clusters.yaml`: new `pts_literature` cluster; removed `etl_literature` cluster.
- `pts.yaml`: 6 new step blocks under `steps:`.
- `unified_pipeline.yaml`: 6 new step entries with `depends_on`; removed `etl_literature` entry; rewired `pts_evidence_postprocess_europepmc.depends_on` to point at `pts_literature_cooccurrence_evidence`.
- `etl.conf`: removed the now-unused `literature: { ... }` block.

### Test plan

- [x] `uv run ruff check src/` clean
- [x] `uv run pytest tests/` passes
- [x] No remaining references to `etl_literature` anywhere
- [x] `pts_literature_*` steps resolve to the `pts_literature` cluster via longest-prefix match
- [x] DAG parses without import errors (validated locally with Airflow present)

### Prerequisites for landing

- PTS `vh-add-literature` merged + tagged release published.
- `pts_version` field in `unified_pipeline.yaml` (separate file from this PR) to be bumped to the new tag before merge.

### Out of scope

- No end-to-end pipeline run in this PR (handoff only; production validation will be a follow-up).
- The `pts-literature-50-secondary` autoscaling policy is assumed to exist in GCP (it was created during the launcher work).

PR #206, opened 2026-06-08
chore: retire ETL stage from unified pipeline

d0choa:
## Summary

The ETL stage in the unified pipeline DAG has had zero step consumers since #195. This PR retires the orphan configuration, loaders, DAG stage function, and supporting operator/enum entries.

Targets the `2606-public-release` branch (PR #201) so it lands together with the public release work.

### Removed
- **`clusters.yaml`**: `etl` and `etl_literature` clusters, plus the `step_job_properties.etl` block.
- **`unified_pipeline.yaml`**: `etl_version` and the `etl_literature` step entry.
- **`etl.conf`**: deleted entirely.
- **`config/unified_pipeline.py`**: the `etl` AppConfig loader (and its PPP overlay), `etl_version` / `etl_jar_origin_uri`, the now-unused `jar_uri()` helper, and the `exts` map in `config_uri()`.
- **`dags/unified_pipeline.py`**: the `etl_stage()` function, its call site, and the imports that only it used (`ETLJobBuilder`, `CopyBlobOperator`, `to_hocon`).
- **`operators/dataproc.py`**: `ETLJobBuilder` class.
- **`models/step.py`**: `UnifiedPipelineStage.ETL` enum entry.
- **`operators/diff.py`**: refreshed docstring examples that referenced the removed `etl_stage` task IDs (now use `pts_target`).

Net diff: **+8 / -410** across 8 files.

## Test plan
- [x] `uv run ruff check src/ tests/` — passes
- [x] `uv run ruff check .` (excluding untracked locals) — passes
- [x] `uv run deptry .` (excluding untracked locals) — passes
- [x] `uv run pytest tests/` — 161 passed, 1 xfailed
- [ ] Render the `unified_pipeline` DAG in Airflow to confirm it loads without the ETL stage


DSuveges (approved):
Thank you for cleaning the orchestration from the scala etl specific logic and configuration. 

- Removal of etl config
- Removal of etl steps from the unified pipeline
- Removal of etl/etl literature pipeline definitions.
- Removal of logic on how etl steps should be called

PR #216, opened 2026-06-30
feat: extend secret model to support JSON-blob secrets

project-defiant:
## Summary

- Add `SecretBlobs` model for GCP secrets whose values are already JSON blobs — written verbatim to `/var/run/secrets/{secret_id}` rather than wrapped in a `{"ENV_VAR": "value"}` envelope
- `SecretInitAction` now accepts `secrets`, `secret_blobs`, or both; a `@model_validator` enforces at least one is set
- `CustomClusterConfig` gains a `secret_blob_list` field; validators and `CreateClusterOperator` updated to handle both secret types
- Add `docs/secrets/README.md` documenting Dataproc and Google Batch secret injection mechanisms

## Motivation

Some secrets in GCP Secret Manager (e.g. the `decode` secret) are stored as JSON blobs. The previous model only supported plain-string secrets via env-var envelope wrapping, which corrupts JSON content.

## Test plan

- [x] Unit tests for `SecretBlobs` model validation
- [x] Unit tests for `SecretInitAction._to_script_str` with blobs-only, secrets-only, and mixed configurations
- [x] Unit tests for `CustomClusterConfig` validation (`secret_blob_list` set without `secret_init_action_uri` raises)


xyg123 (approved):
A bit over my head but if needed for decode I guess we need it, anyways I have faith in the tests, great!

